PR #1, opened 2011-04-10
Added Ability to set $_properties via array key also.

jondavidjohn:
I'm working on a bigger feature in which I need to declare the ORM models like this

```
class Model_Article extends Orm\Model {
    protected static $_properties = array('id' => 'integer', 'title' => 'string', 'contents' => 'text' , 'publish' => 'int[1]');
}
```

I don't believe this change will disrupt any of the current functionality.

Also, not really sure why commit 3260ecf is in there...

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -298,17 +298,17 @@ class Model implements \ArrayAccess, \Iterator {
 		// Return Query object
 		if (is_null($id))
 		{
-			return Query::factory(get_called_class());
+			return static::query();
 		}
 		// Return all that match $options array
 		elseif ($id == 'all')
 		{
-			return Query::factory(get_called_class(), $options)->get();
+			return static::query($options)->get();
 		}
 		// Return first or last row that matches $options array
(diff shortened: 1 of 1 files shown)


jschreuder:
This is already possible but should look like this:

```
protected static $_properties = array('id' => array('type' => 'integer'), 'title' => array('type' => 'string')); // etc
```

This is meant to work with the output of DB::list_columns(), something that wasn't quite finished for RC1 as the Observer_Typing works with "varchar(25)" instead of array('type' => 'string', 'max_length' => 25).

Commit 3260ecf is in there because you are pull-requesting develop branch to master, pull requests should always go to develop anyway and won't ever be accepted on master. The develop branch is always slightly ahead, unless we've just released.

jondavidjohn:
Great! so just grab the current master of the ORM package and those properties definitions should work?

jschreuder:
Only if you would add the Observer_Typing, but that one is going into a rewrite to get it to work with the DB::list_columns() output as I mentioned above.

PR #14, opened 2011-04-25
Chainable Model_Query::select()

crynobone:
Model::find()->select() to return Model_Query object (allow it to be chainable).


jschreuder:
Pulled.

PR #16, opened 2011-04-25
In find method, pass $options to query object even if $id is null.  

santiagofs:
Does no harm if you use find method as usually and helps if you overrides the find method in your models:

namespace Content;

class Model_page extends \Content\Model_content {

public static function find($id = null, array $options = array())
{
$where[] = array('module','=','pages');
array_key_exists('where', $options) and $where = array_merge($options['where'],$where);
$options['where'] = $where;
return parent::find($id,$options);
}

}

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -298,7 +298,7 @@ class Model implements \ArrayAccess, \Iterator {
 		// Return Query object
 		if (is_null($id))
 		{
-			return static::query();
+			return static::query($options);
 		}
 		// Return all that match $options array
 		elseif ($id == 'all')


tmm1:
Hey there,

Soon after the new GitHub Merge Button feature was released today we discovered a bug in the merge procedure in certain cases. This pull request was affected by the bug, causing code to not actually get merged even though it looks from the history as though it was.

To bring the `develop` branch back to its expected state, simply commit this change again:

``` diff
diff --git a/classes/model.php b/classes/model.php
index f493b72..ad827b2 100644
--- a/classes/model.php
+++ b/classes/model.php
@@ -298,7 +298,7 @@ class Model implements \ArrayAccess, \Iterator {
        // Return Query object
        if (is_null($id))
        {
-           return static::query();
+           return static::query($options);
        }
        // Return all that match $options array
        elseif ($id == 'all')
```

Sorry for the inconvenience. If you have any questions or need help, do not hesitate to contact [support](http://support.github.com/).

santiagofs:
Jelmer, do I have to do something with this?

jschreuder:
No, I'll pull manually from your repo - will be fine.

tmm1:
@jschreuder: A pull might not work, as git thinks those commits are already in `develop` even though they were overwritten by the merge commit. You will likely need to make the same change again as a new commit.

jschreuder:
Thanks for the heads up, just figured that out and committed it manually.

PR #28, opened 2011-05-05
Fix Ternary operator result order in Orm / Observer / createdat & updatedat

kurtmilam:
Details: http://fuelphp.com/forums/topics/view/1425

This is my first pull request on GitHub. Hope I didn't screw anything up.

If so, please be gentle :-)

### DIFF
--- a/classes/observer/createdat.php
+++ b/classes/observer/createdat.php
@@ -21,8 +21,8 @@ class Observer_CreatedAt extends Observer {
 
 	public function before_insert(Model $obj)
 	{
-		$obj->{static::$property} = static::$mysql_timestamp ? \Date::time()->format('mysql') : \Date::time()->get_timestamp();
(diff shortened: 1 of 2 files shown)


classes/observer/createdat.php
```diff
@@ -21,8 +21,8 @@ class Observer_CreatedAt extends Observer {
 
 	public function before_insert(Model $obj)
 	{
-		$obj->{static::$property} = static::$mysql_timestamp ? \Date::time()->format('mysql') : \Date::time()->get_timestamp();
```
kurtmilam, line 24: returns \Date::time()->format('mysql') if static::$mysql_timestamp==true - seems to be incorrect to me.

kurtmilam:
Ah, now I think I'm wrong. I misintepreted the meaning of the property name, so I'm going to close this request.

PR #29, opened 2011-05-11
multiple database support

lekoala:
basic support for multiple database

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -36,6 +36,11 @@ class Model implements \ArrayAccess, \Iterator {
 	// protected static $_many_through;
 
 	/**
+	 * @var  string  database
+	 */
+	protected static $_database = null;
+
+	/**
 	 * @var  array  name or names of the primary keys
 	 */
 	protected static $_primary_key = array('id');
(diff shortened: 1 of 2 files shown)


jschreuder:
This looks good, I'll have to fully review & test it but I think you've pretty much covered it.

I have 2 small points already:
- it should be "connection" instead of "database" (because it uses another db connection, not another db)
- when relating models with different connections an exception should be thrown immediately

jschreuder:
I only just noticed: you made the pull request to master, we don't accept requests to master - only to develop. Could you make another pull-request to the correct branch?

patie:
i have suspicion that some problem with $this->connection in omr\classes\query.php because when i set   \
public static $_connection = 'connection2';
model still want working with default connection1 - for example when i call ::find()

jschreuder:
Try using some caps & punctuation next time, that should make that collection of words more like actual sentences.

You're right, there was an ommission in my commit. It's been fixed.

patie:
sry but my english is poor and in destination i use slashes but was be automatically striped

PR #30, opened 2011-05-13
Model Keys

tarnfeld:
I was working on a generic model today for logging stuff, and I wanted a way to know where the log came from - not just the class but the actual row.. I could just insert the ID but since its a generic logging system it could be from any class.

With these two new methods, it generats a "model key" for example Model_Thing_494 and you can call a static method on the \Orm\Model to get back an instance of Model_Thing that has the row with ID 494 loaded up.


WanWizard:
Hmm... Not sure this should be core functionality.

And please don't send pull requests on master. We only accept pull request on the develop branch, to keep merging efforts to a minimum.

jschreuder:
I'm closing this because it's on the master branch. But I think I agree with WanWizard, it's a little specific and will be clutter for most people. We can't include everything or these classes will become too massive.

Also, for your own usage of this, you should use `static::primary_key()` to fetch the primary keys you include with your string representation instead of assuming "id".

tarnfeld:
Yeah, fair enough - just thought i'd see if you guys wanted it :)

Will make sure I make pull requests on the dev branch next time!

@jschreuder Thanks for the advice!

PR #41, opened 2011-05-31
cascade_delete was replaced with cascade_save

davidia:
hasmany ignored cascade_delete configuration and replaced with cascade_save


jschreuder:
It's a bad bug but it's fixed already in develop, we don't accept pull-requests on master.

jschreuder:
Oops, wasn't fixed in develop - not sure why as I'm pretty sure I fixed this before. I'll commit a fix on develop asap.

PR #42, opened 2011-05-31
HasMany cascade_delete configuration ignored and cascade_save used instead

davidia:


jschreuder:
As said in previous pull-request, I've done it immediately - the problem wasn't limited to hasmany.

PR #52, opened 2011-06-28
If the fieldset is loaded from an instance, readding the fields causes an

### LINKED ISSUE #50: With the Validation oberver, can't have multiple objects of same type.
If you have multiple objects of the same type in a file, while that object uses the Validation Observer you get errors like:

```
<strong>Notice:</strong> Field with this name exists already, cannot be overwritten through add(). in PKGPATH/orm/classes/observer/validation.php [43]: add
```

Example how to produce error:

```
$model1 = My_Model::factory(array(...));
$model1->save();

$model2 = My_Model::factory(array(....));
$model2->save();
```

This is caused by Observer\Validation creating it's Fieldset named after the Model's Class. Therefore when you have 2 of the same type of models it ends up using the same Fieldset Instance causing things to muck up.

jeffutter:
If the fieldset is loaded from an instance, readding the fields causes an error. If the field already exists in the fieldset, don't add it again. Fixes #50

I'm not sure this is the most elegant fix, but it works.

### DIFF
--- a/classes/observer/validation.php
+++ b/classes/observer/validation.php
@@ -40,7 +40,7 @@ class Observer_Validation extends Observer {
 
 		foreach ($properties as $p => $settings)
 		{
-			$field = $fieldset->add($p, ! empty($settings['label']) ? $settings['label'] : $p);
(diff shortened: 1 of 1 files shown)


jeffutter:
Ack! i set it to fuel:develop but github didn't listen

PR #54, opened 2011-07-01
Bug fix and feature addition to Validaiton observer

phil-lavin:
Bug fix concerns using related fields on a model which is being validated with the validation observer. When this code executes...

<code>
foreach ($input as $k => $v)
{
    $obj->{$k} = $val->validated($k);
}
</code>

$val->validated('related_field_name') returns false as the related field is not known to the validation object.

The fix adds related fields to the object to ensure validation doesn't replace the array with false. Validation of related fields would be handled using observers in the related model classes.

The feature addition allows a user to do $model->observe('validate') to manually initiate validation. This is useful on the odd occasion you wish to validate without saving.

### DIFF
--- a/classes/observer/validation.php
+++ b/classes/observer/validation.php
@@ -25,8 +25,9 @@ class Observer_Validation extends Observer {
 	 * @param   Fieldset|null
 	 * @return  Fieldset
 	 */
-	public static function set_fields($class, $fieldset = null)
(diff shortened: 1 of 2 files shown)


phil-lavin:
Early duplicate of https://github.com/fuel/orm/pull/59

Closing.

PR #59, opened 2011-07-19
Observer_Validation Changes

phil-lavin:
This is the product of a month or so of work and testing to make this observer fit for general purpose. We are using it (repackaged) internally now.

It's probably not perfect but it can likely be improved over time.

The fixes are, from what I recall, the following...

:: Pass an object instead of a class to allow callables to be added from it to allow for custom validation rules to be added to a model

:: Pull request from https://github.com/fuel/orm/pull/53 has been added to fix what was a fairly fundamental stumbling point in getting it to work

:: Created a validate() method to allow validation without saving

Thoughts?

### DIFF
--- a/classes/observer/updatedat.php
+++ b/classes/observer/updatedat.php
@@ -19,7 +19,10 @@ class Observer_UpdatedAt extends Observer {
 
 	public function before_save(Model $obj)
 	{
-		$obj->{static::$property} = static::$mysql_timestamp ? \Date::time()->format('mysql') : \Date::time()->get_timestamp();
+		if ($obj->is_new() || $obj->is_changed())
+		{
+			$obj->{static::$property} = static::$mysql_timestamp ? \Date::time()->format('mysql') : \Date::time()->get_timestamp();
+		}
 	}
 }
 
--- a/classes/observer/validation.php
+++ b/classes/observer/validation.php
@@ -25,8 +25,9 @@ class Observer_Validation extends Observer {
 	 * @param   Fieldset|null
 	 * @return  Fieldset
 	 */
-	public static function set_fields($class, $fieldset = null)
+	public static function set_fields($obj, $fieldset = null)
 	{
+		$class = get_class($obj);
 		$properties = $class::properties();
 
 		if (is_null($fieldset))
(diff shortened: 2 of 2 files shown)


jschreuder:
It's moving in the right direction, but there's a couple of things that aren't right yet.

First the tip also shared on IRC: you can just do `$obj->properties()` instead of `$class = get_class($obj); $class::properties()`, calling it on the instance is a lot faster as the `$class::properties()` is basicly the same as `call_user_func(array($class, 'properties'))` and that's slow.

One of the major points here is about re-usability, which becomes a problem in the current code because the fields are always added. The solution from pull-request #50 isn't elegant as I said, and it still isn't. Fields should only be added once, not iterated over again and again - once the fieldset is generated it should just do.  
This problem is further demonstrated by the right idea to add the object as a callable, but now every time validation is run another object is added to the callables. If you have 100 objects to validate, there's a 100 callables added at the end.  
The bottom line here is that we need a solution to detect whether the fields were added to the Fieldset instance already. Best solution I can think of right now is to add a static variable to the method that keeps track of the fieldsets it has created already, something like this:

```
// at the top of the method
static $generated = array();

// after the fieldset was either fetched or created
! array_key_exists($class, $generated) and $generated[$class] = array();
if ( ! in_array($fieldset, $generated[$class], true))
{
    // add_callable() & add the fields non-conditional

    // end with this to prevent regeneration
    $generated[get_class($obj)][] = $fieldset;
}
```

Also as discussed on IRC the fix for relations is wrong. Those shouldn't be there as empty fields. The problem lies in the fact that when this was created the ArrayAccess didn't return relations, because it does now the relations are added to the $input array in the validate() method while they shouldn't be. That's also where both the problem and the solution lies: Relations shouldn't be added to the input array.

And lastly: the [Model::set_form_fields()](https://github.com/fuel/orm/blob/master/classes/model.php#L1101) method must become non-static and pass $this instead of get_called_class().

I agree with the validate() method, that's a good idea.

jschreuder:
As you haven't responded yet I'm going to assume this will do and close this issue.

phil-lavin:
Haven't had chance to try it. Likely won't get chance for a few weeks. I'll reopen if/when I get chance to test.

PR #65, opened 2011-08-03
Added support for hydration via DB::as_model

frankdejonge:
So we can do:

$model = DB::select()->from('table')->as_object('Model_Something')->exectute()->current();

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1061,7 +1061,7 @@ class Model implements \ArrayAccess, \Iterator {
 		$property = (array) $property ?: array_keys(static::properties());
 		foreach ($property as $p)
 		{
-			if ($this->{$p} !== $this->_original[$p])
+			if ( ! isset($this->_original[$p]) or $this->{$p} !== $this->_original[$p])
 			{
 				return true;
 			}
--- a/classes/model.php
+++ b/classes/model.php
@@ -543,6 +543,12 @@ class Model implements \ArrayAccess, \Iterator {
 	 */
 	public function __construct(array $data = array(), $new = true)
 	{
+		if( ! empty($this->_data))
+		{
+			$this->_original = $this->_data;
+			$new = false;
+		}
+		
 		if ($new)
 		{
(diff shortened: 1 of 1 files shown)


classes/model.php
```diff
@@ -543,6 +543,20 @@ class Model implements \ArrayAccess, \Iterator {
 	 */
 	public function __construct(array $data = array(), $new = true)
 	{
+		if( ! empty($this->_data))
+		{
```
jschreuder, line 547: @FrenkyNet I don't get this, how could this ever happen? When `__construct()` is called it's always going to be empty...?

classes/model.php
```diff
@@ -1055,7 +1069,7 @@ class Model implements \ArrayAccess, \Iterator {
 		$property = (array) $property ?: array_keys(static::properties());
 		foreach ($property as $p)
 		{
-			if ($this->{$p} !== $this->_original[$p])
+			if ( ! isset($this->_original[$p]) or $this->{$p} !== $this->_original[$p])
```
jschreuder, line 1072: Why is this? To prevent a PHP notice?

Becasue basicly it's wrong, the lack of a value doesn't signify whether or not it's been changed. Also `isset()` returns false when the value is null, which it might very well be set to.

frankdejonge, line 1072: this should be array_key_exists probalby. it's for supporting subsets.

classes/model.php
frankdejonge, line 547: I'll setup an example so you can see how it works.

jschreuder, line 547: That's just weird, but good to know.

classes/model.php
jschreuder, line 1072: There will be more problems for supporting subsets, allowing for that won't be as simple as this. I still haven't gotten round to checking up on that.

As to this part of any eventual solution: now you are just ignoring any values that weren't there when loading but are there when saving. It should either be saved or throw an exception when being set. Allowing to set it but not saving it is wrong.

PR #68, opened 2011-08-09
Nested where causes an error

davidia:
Added a check to avoid undefined element notice


jschreuder:
It's not yet up to coding guidelines: the `!` must have a space on both sides.

PR #81, opened 2011-09-06
Flexibility with the way you use the Orm without impacting performance

bencorlett:
Earlier today I spoke to @dhorrigan about chaining in the Session class. It got me thinking about the Orm.

Instead of setting properties each time, sometimes it is nice to have methods to set the properties, for two reasons:
1. It allows chaining
2. It allows you to override these methods in your model and manipulate the results before returning them.

After building this I realised a feature lacking from the Orm that Kohana's Orm has - a values() method - to reassign data to a model from an array - once again saving on typing.

This screenshot below demonstrates the enhanced usage from the attached commits:
http://d.pr/FcnR

And this one below demonstrates a huge advantage gained by using the convenience methods:
http://d.pr/uCMs

I'm aware that __call() has some performance implications, however as you'll see you don't have to use methods requiring __call(), but it adds a whole lot of power. The best part? If people don't want to use the Orm this way it won't affect them at all! This feature is needed IMO especially because the Orm is not designed to be easily overridden.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -564,18 +564,7 @@ class Model implements \ArrayAccess, \Iterator {
 
 		if ($new)
 		{
-			$properties = $this->properties();
-			foreach ($properties as $prop => $settings)
-			{
-				if (array_key_exists($prop, $data))
-				{
-					$this->_data[$prop] = $data[$prop];
-				}
-				elseif (array_key_exists('default', $settings))
-				{
-					$this->_data[$prop] = $settings['default'];
-				}
-			}
+			$this->values($data);
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


jschreuder:
Basicly I think this is fine but I have a few points:
- `uns()` must go, it should be `unset()` if it would go in but the truth is that it's  useless. You can't unset something and the `__unset()` method is there just to allow `unset($model->relation)` not because that actually makes sense for properties (set it to `null` yourself).
- is `values()` only for initial hydration? Otherwise defaults will overwrite existing values if you don't pass them, which seems bad.
- The bodies of `__get()` and `__set()` should be moved to get() & set() and the magic method should use them. That's because using magic methods directly is notoriously slow for some weird reason.

bencorlett:
Hi,

I'll reply in order :)
- I couldn't get a method called unset() to work, it was saying unexpected string expecting function or whatnot, i'm assuming it's a reserved word?
- The values() method, I just realised should just loop through the array posted and set the property values, I just copied that code from __construct() and made reference to it. I'll fix that up in the morning and I'll put in another pull request.
- Okay I can change that over NP

jschreuder:
Yeah, unset is probably reserved (the unset() function isn't actually a function but a language construct). Still has to go though.

Don't forget to keep hydration with defaults for the constructor.

bencorlett:
Not a problem. I'll have something for you very soon.

Cheers mate

bencorlett:
Hi @jschreuder, I've added another commit - hopefully that fixes it.

ghost:
Yay! No need to use Kohana ORM in fuel projects anymore

PR #86, opened 2011-09-16
1.1/develop I've added a Slug Observer

ushis:
The slug observer creates a unique slug for a model using Inflector::friendly_title() and adds an index, if the slug already exists.

Observer_Slug::$source is the property, which is give to Inflector::friendly_title()
Oberserver_Slug::$property is the property for the slug

Maybe this is somethong for you.


ushis:
Damn, wrong branch. sorry

PR #87, opened 2011-09-16
1.1/develop Added a Slug Observer

ushis:
The slug observer creates a unique slug for a model using Inflector::friendly_title() and adds an index, if the slug already exists.

```
Observer_Slug::$source   // is the property, which is given to Inflector::friendly_title()
Observer_Slug::$property // is the property for the slug
```

Maybe this is something for you.

### DIFF
--- a/classes/observer/slug.php
+++ b/classes/observer/slug.php
@@ -34,7 +34,7 @@ class Observer_Slug extends Observer {
 	{
 		$class = get_class($obj);
 		$slug  = \Inflector::friendly_title($obj->{static::$source}, '-', true);
-		$same  = $class::find()->where(static::$property, 'like', $slug.'%')->get();
+		$same  = $class::find()->where(static::$property, 'regexp', '^'.$slug.'(-[0-9]+)?$')->get();
 
 		if ( ! empty($same))
 		{
--- a/classes/observer/slug.php
+++ b/classes/observer/slug.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Fuel is a fast, lightweight, community driven PHP5 framework.
+ *
+ * @package		Fuel
+ * @version		1.1
+ * @author		Fuel Development Team
+ * @license		MIT License
+ * @copyright	2010 - 2011 Fuel Development Team
+ * @link		http://fuelphp.com
+ */
+
+namespace Orm;
+
+class Observer_Slug extends Observer {
+
+	/**
+	 * @var string Source property, which is used to create the slug
+	 */
+	public static $source = 'title';
+	
+	/**
+	 * @var string Slug property
+	 */
+	public static $property = 'slug';
+
+	/**
+	 * Creates a unique slug and adds it to the object
+	 *
+	 * @param  Model The object
+	 * @return void
+	 */
+	public function before_insert(Model $obj)
+	{
+		$class = get_class($obj);
+		$slug  = \Inflector::friendly_title($obj->{static::$source}, '-', true);
+		$same  = $class::find()->where(static::$property, 'like', $slug.'%')->get();
+
+		if ( ! empty($same))
+		{
+			$max = 0;
+			
+			foreach ($same as $record)
+			{
+				$index = (int) preg_replace('/^[^\n]+-([0-9]+)$/', '\1', $record->slug);
+				$max < $index and $max = $index;
+			}
+			
+			$slug .= '-'.($max + 1);
+		}
+			
+		$obj->{static::$property} = $slug;
+	}
+}
--- a/classes/observer/slug.php
+++ b/classes/observer/slug.php
@@ -42,7 +42,7 @@ class Observer_Slug extends Observer {
 			
 			foreach ($same as $record)
 			{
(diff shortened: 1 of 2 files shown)


jschreuder:
The code is good but I'm pondering whether or not I think it should be in here. Some code just has more of a place as third-party code because it's too much implementation - all the other included observers are very generic, this would be the first one specificly for a blog/cms purpose.

One bug in the logic though if I'm not mistaken. Let's say I add a page called "Slugtest", that would go into the DB with a slug "slugtest". After that I add a page called "Slug", it creates the slug "slug" and does the search using "slug%" and finds my "slugtest". preg_replace return the full string "slugtest" which is cast to int and reported as 0. And my first page with slug "slug" is slugged to "slug-1".  
You should still check whether the found slug actually fully matches the slug you're trying to number and only change the max if it does.

ushis:
Yes, you're right, The code is quite specific, but i think slugs are very common and if my framework of choice supports it out of the box, i would be very happy. But it is your decision.

You are right with the bug. I'll think about that.

classes/observer/slug.php
```diff
@@ -0,0 +1,54 @@
...
+	 * Creates a unique slug and adds it to the object
+	 *
+	 * @param  Model The object
+	 * @return void
+	 */
+	public function before_insert(Model $obj)
+	{
+		$class = get_class($obj);
+		$slug  = \Inflector::friendly_title($obj->{static::$source}, '-', true);
+		$same  = $class::find()->where(static::$property, 'regexp', '^'.$slug.'(-[0-9]+)?$')->get();
```
jschreuder: You can call `find()` on the $obj instance, that's much faster as this is effectively the same as using `call_user_func()` which is pretty slow. Thus use `$obj->find()` instead.

Also I'm not sure about having SQL do the REGEXP, is that widely supported by non-mysql implementations? I know the ORM is pretty much mysql only right now, which is something we intend to fix and don't want to introduce more mysql-only code.

ushis:
I'm not sure about the REGEXP support too. I'll change it...

ushis:
Ok, i replaced the REGEXP with LIKE and added a check in the loop. I think this should work.

But i'm not very happy about the 2 preg_ functions. i'll think about that.

Hmm... What do you think about something like that:

```
foreach ($same as $model)
{
    $index = (int) preg_replace('/^'.$slug.'-([0-9]+)$/', '\1', $model->{static::$property});
    $index == 0 and $model->{static::$property} != $slug and $index = -1;           
    $max < $index and $max = $index;
}
```

jschreuder:
Use the third param of preg_match() (check [php.net docs](http://nl3.php.net/manual/en/function.preg-match.php)).

Something like this:

``` php
<?php
if (preg_match('/^'.$slug.'-([0-9]+)?$/', $record->{static::$property}, $matches))
{
     $index = (int) $matches[1];
     $max < $index and $max = $index;
}
```

jschreuder:
I've decided to merge it once you've fixed this last thing, but before I do could you also add a section about this observer to the docs: https://github.com/fuel/docs/blob/1.1/develop/packages/orm/observers/included.html

ushis:
Oh so easy... There is a little bug in your code. I'll use this

```
<?php
if (preg_match('/^'.$slug.'(?:-([0-9]+))?$/', $model->{static::$property}, $matches))
```

This pattern will also match non-indexed slugs

ushis:
Ok, I've added a section to the docs: https://github.com/fuel/docs/pull/196

frankdejonge:
Just a thought. How about using an array for the propperty so in that order the slug will be generated array('id','name') will generate 326-my-article-name. Makes it less app specific, and more usable. @jschreuder: agree?

jschreuder:
@FrenkyNet it's not really app-specific now is it? You want to allow including the id in which case adding an index isn't necessary anymore, I could see that as an option, though on the other hand it's not really necessary to save it in the db if your URL is always `$id.'-'.Inflector::friendly_title($title)` in that case you don't really need to save it, fetching the ID from the URL is enough to fetch the page and if you want to enforce the title you can just check the rest against `Inflector::friendly_title($title)`. Saving it to the DB in such a case is basicly adding unecessary redundancy IMO.

frankdejonge:
@jschreuder: Deciding the slug is the url friendly version of 1 field for me is an app-specific decision. Lots of people have their own slug setup including multiple fields to keep them unique. The id is just one of those options. And you're right that if a user decides to have the id in the slug the easier way to get it out would be to strip is of and look for that, but there are other things. Like with user accounts a slug might be generater by ('first_name', 'last_name'). By not having the id and no clear way of knowing where the last name starts, the full slug will be used in the search.

jschreuder:
@FrenkyNet ok, that's a pretty easy fix: allow the `$property` to be both a string and an array. Use it by doing something like this: $slug_properties = (array) static::$property; and foreach the properties with dashes in between.

This is going to get us into trouble btw as setting this by static propety has been a bad practice from the beginning as the setting needs to be model-specific. I'm hoping I'll find the time to change this before 1.1.

PR #92, opened 2011-09-21
orm has_one relation bug fix

kapv89:
when i try to delete an object of a model(say model_1) which has a has_one relation with another model(say model_2), and if no one-to-one record for that object of model_1 exists, i get an error:
 "Call to a member function frozen() on a non-object".
In order to fix this, I have modified line 188 to:
if ($model_to and ! $model_to->frozen())
as this first checks whether "$model_to" exists or not and then calls the "frozen()" function.

More detail can be found here: http://fuelphp.com/forums/topics/view/4888

### DIFF
--- a/classes/hasone.php
+++ b/classes/hasone.php
@@ -185,7 +185,7 @@ class HasOne extends Relation {
 		$model_from->_relate($rels);
 		$model_from->freeze();
 
-		if ( ! $model_to->frozen())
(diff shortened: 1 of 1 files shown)


jschreuder:
Still on the wrong branch but I pulled it using another method, remember to select the 1.1/develop branch next time (it defaults to 1.0/master).

kapv89:
O.K.

Issue #101, opened 2011-10-07
\Orm\ValidationFailed would be really great if it contained the validation instance


phil-lavin:
It was a great change to include the string output of errors from the validation in the message of the exception. Would it be worth my while adding a reference to the whole Validation instance to this exception also? It'd be really useful, especially with Orm validation, to be able to obtain this without having to obtain a fieldset instance of the same name as the model the validation was run on.

If you think this is worthwhile, I'll submit a patch.

jschreuder:
That could be usefull, though reference the fieldset and not the validation (you can get the validation from the fieldset).

phil-lavin:
Ok, will do on Monday. Thanks :)

PR #102, opened 2011-10-12
Related #101

phil-lavin:


jschreuder:
Looks good but can you add docblocks?

phil-lavin:
Is that ok or do you need them to be a little more descriptive of the params?

jschreuder:
That's ok for now.

PR #103, opened 2011-10-13
Added support for multiple source properties to Observer_slug

ushis:
Now you can use

``` php
<?php
Orm\Observer_slug::$source = array('name', 'favourite_color');

// Creates a slug like 'george-blue-3'
```

like discussed here: https://github.com/fuel/orm/pull/87#issuecomment-2303133


tarnfeld:
Oooh, this is a pretty awesome feature to have in the core :D I wrote my own thingy for doing slugs but this is way better!

PR #104, opened 2011-10-13
New data_type 'date' for typ

anteater-zz:
I added a new `data_type` `date` for typing. This gives you the possibility to retrieve your date/time related model-fields in a format you choose. It works with the `\Fuel\Date` object for i18n.
This is a short sample of how to use it.

```
class Model_Project extends Orm\Model
{
    protected static $_table_name = 'projects';
    protected static $_primary_key = array('id');

    protected static $_properties = array (

        'id'            => array('data_type' => 'int'),
        'start'         => array('data_type' => 'date', 'format_model' => 'timestamp', 'format_view' => 'us'),
        'end'           => array('data_type' => 'date', 'format_model' => 'timestamp', 'format_view' => 'us')

    );
}
```

You can choose any format which is defined in the `date.php` config file, either in `CORE` or `APP` path. `format_model` defines the pattern on the persistance layer (db), while `format_view` is the pattern of the final represenation on your view.
Have fun, or throw away, if it is not needed :)

Oh, and sorry for the mess. GitHub beginner.

### DIFF
--- a/classes/observer/typing.php
+++ b/classes/observer/typing.php
@@ -58,6 +58,10 @@ class Observer_Typing {
 			'before' => 'Orm\\Observer_Typing::type_json_encode',
 			'after'  => 'Orm\\Observer_Typing::type_json_decode',
 		),
+		'/^date$/uiD' => array(
+			'before' => 'Orm\\Observer_Typing::type_date_decode',
+			'after'  => 'Orm\\Observer_Typing::type_date_encode',
+		),
 	);
 
 	/**
@@ -304,6 +308,31 @@ class Observer_Typing {
 	public static function type_json_decode($var)
 	{
 		return json_decode($var);
+	}
+	
+	/**
+	 * Return a Fuel Date object on the base of a given format
+	 *
+	 * @param int value
+	 * @param array
+	 * @return \Fuel\Core\Date
+	 */
+	public static function type_date_encode($var, $settings)
+	{
+		$pattern = isset($settings['format']) ? $settings['format'] : 'local';
+		return \Date::create_from_string($var , $pattern);
+	}
+	
+	/**
+	 * Return a formated value of a given Fuel Date object
+	 *
+	 * @param \Fuel\Core\Date
+	 * @return int
+	 */
+	public static function type_date_decode(\Date $var, $settings)
+	{
+		$pattern = isset($settings['format']) ? $settings['format'] : 'local';
+		return $var->format($pattern);
 	}
 }
 
--- a/classes/observer/typing.php
+++ b/classes/observer/typing.php
@@ -15,9 +15,10 @@ namespace Orm;
 // Invalid content exception, thrown when conversion is not possible
 class InvalidContentType extends \UnexpectedValueException {}
 
-class Observer_Typing
-{
+class Observer_Typing {
 
+	public static $date_format = 'mysql';
+	
 	/**
 	 * @var  array  types of events to act on and whether they are pre- or post-database
 	 */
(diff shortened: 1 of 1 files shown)


jschreuder:
This doesn't follow our coding standards: http://fuelphp.com/docs/general/coding_standards.html

Other changes required for acceptance:
- Don't decode to a date, decode to the Date object so you can still manipulate it afterwards (allowing you to get rid of the `format_view` option)
- Defualt to UNIX timestamp and allow a switch like the CreatedAt/UpdatedAt observers use: "mysql_timestamp" (which thus defaults to false)

With those changes it's more consistent with the other classes and won't need exceptions thrown with underconfiguration.

sheeep:
Okay, 
- I managed to get rid of the `format_*` stuff by taking just a `static::date_format`
- Default is (like in CreatedAt) UNIX timestamp.

But: We still need an exception with underconfigurated systems, as `\Date` would end up in a notice ("Input was not recognized by pattern.") and screw the content of the field.

I don't get what you mean by "doesn't follow our coding standards..". Does it now?

ushis:
@sheeep Following the fuel coding standards means to change things like this:

``` php
<?php
// Your style
if(!$ret = \Date::create_from_string($var, static::$date_format))

// Fuel style
if ( ! $ret = \Date::create_from_string($var, static::$date_format))
```

You can find the standards here: http://fuelphp.com/docs/general/coding_standards.html#control_structures

And the new (only new class braces) standards: http://fuelphp.com/dev-docs/general/coding_standards.html

sheeep:
@ushis: I changed the mentioned spaces around the `!` and re-added the newline before the first `{`.
The dev-docs indicates that you have to add (or in our case keep) another newline after the `{`, but I'm not sure. Maybe that would be a point to add.

classes/observer/typing.php
```diff
@@ -59,6 +61,10 @@ class Observer_Typing
 			'before' => 'Orm\\Observer_Typing::type_json_encode',
 			'after'  => 'Orm\\Observer_Typing::type_json_decode',
 		),
+		'/^date$/uiD' => array(
```
ushis, line 64: i would change this line to

``` php
'/^date(time)?$/uiD' => array(
```

ushis:
@sheep I think it is ok. @jschreuder has the last word.

**EDIT** But your default format is not UNIX timestamp. It is myqsl.

classes/observer/typing.php
sheeep, line 64: As the comment above this code block is `regexes for db types` you are right. But what about the other Date and Time related types mentioned in the documentation? http://dev.mysql.com/doc/refman/5.0/en/date-and-time-types.html

ushis, line 64: I would support date, datetime, timestamp. These three types are supported by `\Date::create_from_string($date, 'mysql');` But i am not sure about other sql implementations. Do they have such data types?

sheeep:
@ushis Ok, this is embarrassing! Two possibilities:
1) Add a timestamp entry to the date.php configuration file and replace `mysql` with the corresponding key. (`'timestamp' => '%s'`)
2) Rewrite the functions to listen to a setting variable (`format`).

I'd prefer the first way, as still don't get it why there isn't such a key/value pair anyway :)

ushis:
@sheeep why dont do it like in Observer_CreatedAt? Set a property `$mysql_timestamp = false;` And use somthing like:

``` php
<?php
public static function type_date_decode(\Date $var)
{
    return static::$mysql_timestamp ? $var->format('mysql') : $var->get_timestamp();
}
```

And a similar method for encode...

Their is no such key/value pair, because the Date class supports timestamp out of the box: `$date = \Date::forge($timestamp);`

sheeep:
@ushis: This would limit the number of usages. Let's say you just want to store the year of somewhat in you database, you would have to map it down either to mysql-timestamp or UNIX-timestamp. It does make sense in a `Created/UpdatedAt`, but not in this case, I guess.
Or am I wrong?

I do understand that the Date class supports timestamp, but in this exactly case you would have to check for a valid pattern first. If it is, use `Date::create_from_string`, if not use `Date::forge`. This is somehow inflexible.

jschreuder:
@sheeep there's a real problem with `Date::create_from_string()`: it doesn't fully work on Windows, there it uses `strtotime()` instead of `strptime()`.

There's also a best/bad practice POV to take here: there's pretty much only 2 right ways to save date-time info, and that's with MySQL timestamps and UNIX timestamps. People can use other formats if they really want to, but it's not necessary for a framework to support bad practice.

ushis:
@sheeeb Maybe we should move this discussion into the irc...

If i only want to store the year, i dont want to have a Date object as DB result. That makes no sense.

jschreuder:
@ushis in which case you don't use this data_type but just int.

ushis:
@jschreuder That's the point.

jschreuder:
@ushis ok, then we understand each other :)

jschreuder:
I've thought about this and decided to do it differently: there will be 2 types: time_unix & time_mysql which will use the expected timestamps.

PR #108, opened 2011-10-24
type_time_decode should be static

ushis:
@jschreuder i really dont want to be an ass.


jschreuder:
@ushis you're not, I just committed it before dinner with the intend to fully check it once I'm done eating - but you're doing that for me :P

PR #112, opened 2011-10-29
Add before_find and after_find event calls to the find function. Idea from Model_crud pre_find and post_find functions.

rclanan:
https://github.com/fuel/core/blob/1.1/develop/classes/model/crud.php#L197
https://github.com/fuel/core/blob/1.1/develop/classes/model/crud.php#L202

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -379,15 +379,25 @@ class Model implements \ArrayAccess, \Iterator {
 	 */
 	public static function find($id = null, array $options = array())
 	{
+		$this->observe('before_find');
+
 		// Return Query object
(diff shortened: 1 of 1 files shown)


rclanan:
This was meant for a proposal. Hopefully to be looked at for properly being implemented as dhorrigan said this wouldn't actually work as expected. 

dhorrigan: abyssal: the idea of the before and after find is that before you can modify the query and after you can modify the result.
dhorrigan: abyssal: no…it should be after the options are set

Closing this pull request and putting in a feature request for the proposal for this to be implemented by someone who knows what they are doing!

PR #115, opened 2011-10-31
Add support for omitting fields from Fieldsets

kbanman:
Allows you to specify in a model's `$_properties` whether a field should be included/excluded from Fieldsets/Validation.

Example:

```
protected static $_properties = array(
    'id' => array(
        'data_type' => 'int', 
        'form' => array(
            'include' => false,
        ),
    ), 
);
```

(The default is, of course, to include the field)

Use case being that you usually wouldn't want to display fields like `created_at` or `updated_at` in a form.

### DIFF
--- a/classes/observer/validation.php
+++ b/classes/observer/validation.php
@@ -77,6 +77,10 @@ class Observer_Validation extends Observer
 		$properties = is_object($obj) ? $obj->properties() : $class::properties();
 		foreach ($properties as $p => $settings)
 		{
+			if (isset($settings['form']['include']) and ! $settings['form']['include'])
+			{
+				continue;
+			}
 			if (isset($settings['form']['options']))
 			{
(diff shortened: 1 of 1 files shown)


jschreuder:
Give the 'form' array's 'type' key `false` as a value: https://github.com/fuel/core/blob/1.1/develop/classes/fieldset/field.php#L436

kbanman:
Oh, excellent, thanks.

kbanman:
Okay, that method suppresses the actual field from being displayed, but the label is still generated and displayed. I'll see if I can figure it out and submit another pull request.

jschreuder:
Ah damn, I think I messed that up when I moved it to Fieldset_Field from the Fieldset class.

Fix should be easy, from line 444:

``` php
<?php
if (empty($build_field) or $this->type == 'hidden')
{
    return $build_field;
}

return $this->template($build_field);
```

PR #119, opened 2011-11-08
One single Call create()

ronan-gloo:
ActiveRecord-like feature to create new Record.

`$post = Post::create(array('title' => 'yuh', 'body' => 'count'));`

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -964,7 +964,7 @@ class Model implements \ArrayAccess, \Iterator {
 			$this->unfreeze();
 
 			// Insert or update
-			$return = $this->_is_new ? $this->create() : $this->update();
+			$return = $this->_is_new ? $this->_create() : $this->update();
 
 			$this->freeze();
 			foreach($this->relations() as $rel_name => $rel)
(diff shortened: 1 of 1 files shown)


jschreuder:
While probably perfect for you, in general it just confuses things (forge/create) and only to save you the trouble of adding ->save() at the end. This is something more suited for you own extension of the Orm\Model class and has no place in the core (I already dislike having both query()/find() due to BC)

[edit] And as a side-point: it makes the class internally inconsistent by having `_create()` named differently from `update()`.

ronan-gloo:
Well, i understand your point about forge/create confusion, even if 'create' is a well-known ActiveRecord call.
I already use this method in my general model extension, mainly because chaining ->save() after ::forge() returns a bool, and not the object (which follow the internal logic).

about the _ prefix for original create method, in my class extension, i've named this method 'add' (Post::add(...)) to not break ORM\Model class updates. I've just sent you an example, original protected method or the new public can be named 'add'.

jschreuder:
Just to clarify: the main point is that there's a lot of possible handy methods we can add, and we've had pull-requests for all of them. The Orm\Model is already pretty big and if I started allowing shortcuts in it'd grow way too fat in a very short amount of time, this is why I deny most pull-requests like this - a shortcut only makes it fatter without adding functionality.

I'll probably add a note about this to the docs: if you want stuff like that, feel free to extend Orm\Model as that's what I do myself as well to add some app-specific logic and shortcuts to the model. And that's where things like this belong.

ronan-gloo:
Ok, no problems, i understand your point of view.

Issue #120, opened 2011-11-13
\Orm\Query create duplicate order_by statement.


crynobone:
There a bug with the current Orm package for FuelPHP, seem to me the order_by was added twice.
## Reference:
- https://github.com/fuel/orm/blob/1.1/develop/classes/query.php#L614
- https://github.com/fuel/orm/blob/1.1/develop/classes/query.php#L765
## Example Code:

```

\Model_Work::query()
    ->order_by('to', 'ASC')
    ->order_by('from', 'DESC')
    ->get();

# Generate
SELECT ...
FROM `works` AS `t0` 
ORDER BY `t0`.`to` ASC, `t0`.`from` DESC, 
`t0`.`to` ASC, `t0`.`from` DESC
```

WanWizard:
Issue was fixed and merged months ago. Closing issue.

PR #122, opened 2011-11-15
Make sure $order_by is empty after adding it to query builder

crynobone:
Make sure $order_by is empty after adding it to query builder, following code should only cater order_by from relations

(solved #120)
Signed-off-by: crynobone crynobone@gmail.com

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -612,6 +612,9 @@ class Query
 
 		// Get the order
 		$order_by = $this->order_by;
+
+		// create a backup for subquery
+		$order_by_backup = $order_by;
 		if ( ! empty($order_by))
 		{
 			foreach ($order_by as $key => $ob)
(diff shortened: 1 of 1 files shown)


jschreuder:
Did you test this with subqueries and multiple order statements? I think you just broke a hell of a lot...

crynobone:
Updated the commit to use order_by backup if `$this->use_subquery()` return `true`.

# Case Study

```
<?php

class Model_User extends \Orm\Model {

    protected static $_has_many = array(
        'works'           => array(
            'model_to'   => "\Model_Users_Work",
            'conditions' => array(
                'order_by' => array(
                    'from' => 'ASC',
                    'to'   => 'DESC',
                ),
            ),
        ),
    );

}
```

Both before and after changeset return

```
SELECT ... 
FROM (
    SELECT … 
    FROM `users` AS `t0` 
    ORDER BY `t0`.`id` ASC LIMIT 1
) AS `t0` 
LEFT JOIN `users_works` AS `t1` 
ON (`t0`.`id` = `t1`.`user_id`) 
WHERE (`t0`.`username` = 'crynobone') 
ORDER BY `t0`.`id` ASC, `t1`.`from` ASC, `t1`.`to` DESC
```

Using

```
$user = \Model_User::find_by_username('crynobone', array('related' => array('works')));
```

jschreuder:
Thanks.

PR #126, opened 2011-12-12
Added method to clear cached stuff

kapv89:
I recently encountered a situation where after some heavy db operations..... i tried a json_encode($model->to_array), and got an out-of-memory error. The problem was that the orm had cached a lot of stuff and a json_encode had a huge array to traverse. I added a method to clear cached objects, and cleared cached objects(:P), did a query again, and it worked.

Here I have added some static methods to clear stuff which the orm/model caches ...

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -105,6 +105,31 @@ class Model implements \ArrayAccess, \Iterator {
 		'has_many'      => 'Orm\\HasMany',
 		'many_many'     => 'Orm\\ManyMany',
 	);
+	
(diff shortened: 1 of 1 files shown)


jschreuder:
The problem with this is that you allow the same object to exist twice in runtime, the very thing this cache was created to prevent. That will cause problems with updating and cause ambiguity in code.

A better solution for something like this would be to allow you to "refresh" an object already in the cache, thus overwriting the original and edited values already in the cache with newly and less (or more) extensive relations.

PR #137, opened 2012-02-14
Removing select relation columns on count

dbpolito:
There is no reason to select the columns of the relations when you calls the count method.


jschreuder:
This ignores the effect any where conditions upon relations or relation join types might have. It may save a bit on a simple count() but it complicates all those other things.

PR #140, opened 2012-02-24
array of primary keys as a parameter for find method

sahanh:


jschreuder:
That would completey drop support for tables with multiple primary keys.

PR #141, opened 2012-03-06
NULL values are wrongly reported as changed.

Savageman:
Spotted this when using get_diff().


WanWizard:
@jschreuder sounds logical, any comments?

PR #142, opened 2012-03-12
This seems to be a git merging bug (it duplicated a code chunk)

ghost:
This seems to be a git merging bug (it duplicated a code chunk) that sometimes causes errors with cascade deletes :

The $rel->delete(…) is found twice in the Model function delete():
- at lines 1120, before $query->delete() 
- at line 1141 after $query->delete()

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1138,13 +1138,6 @@ class Model implements \ArrayAccess, \Iterator
 				return false;
 			}
 
-			$this->freeze();
-			foreach($this->relations() as $rel_name => $rel)
-			{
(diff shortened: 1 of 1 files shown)


jschreuder:
If you'd have looked closely you'd have noticed they're not entirely the same. Some relation types are processed before the model you're acting on, while others are processed after. Which is why this foreach is ran twice, but none of these is deleted twice as they only work half the time.

ghost:
Oh yes,

Thanks for the quick answer.
Sorry for the wrong diagnostic, I did not see the yes/true. Actually it occurred to me that git duplicated some chunks of code…

So, I need to check elsewhere why I'm having trouble with my cascade deletes.

Dimitri

PR #145, opened 2012-04-05
Change the visibility of properties

felixgilles:
Change the visibility of properties ($_is_new, $_frozen, $_data, $_original, $_data_relations, $_original_relations, $_view) from private to protected so they can be accessed when class is extended


jschreuder:
Fixed by 9b6d28f660ea471292ea3c4d8a80d58def6874c4

PR #148, opened 2012-04-24
Accessing relation on new Model Object bug fix

andrewwheal:
Don't bother trying to fetch relations of a model when get is called and the object is new


peter-fritchley:
This problem still exists, and becomes apparent when unsetting relations, this is because it makes the key_from null in the database for the related record - therefore when attempting to access a relation on a new model, it fetches all of the relations where the corresponding key is null.

WanWizard:
Implemented here: https://github.com/fuel/orm/pull/152

PR #149, opened 2012-04-27
Fix error undefined index @ line 850:

purwandi:
Signed-off-by: Purwandi free6300@gmail.com

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -857,7 +857,7 @@ class Query
 				$s[0] === $prefix.$pk and $pk_c = $s[1];
 			}
 
-			if (is_null($row[$pk_c]))
+			if ( ! isset($row[$pk_c]))
 			{
 				return false;
 			}


purwandi:
When i create query from ORM like this

```
$student = Model_User::find('all',array(
    'related'   => array(
        'user_metadata',
        'user_group'    => array(
                'where' => array(
                    array('group_id',3)
                        )
               )
    )
));
```

I have got error 

http://d.pr/i/QUHL

WanWizard:
First, your error messages shows line 850, while the actual line (and your change) is on line 860. So were you working with an older version?

Second, I'm not sure this is right. If $row[$pk_c] is not set at that moment, that can only mean $pk_c is still null, which means it couldn't find a primary key. Which would point towards a definition error in your model.

@jschreuder, your view?

purwandi:
@WanWizard  ah yes, i working with FuelPHP 1.1

purwandi:
I had to update to 1.1/develop, but I found the same problem when i use `is_null` .

http://d.pr/i/tGvs

WanWizard:
Please post all models involved on http://scrp.at so I can have a look at it. It shouldn't do this if all keys are defined properly.

purwandi:
OK, Here is my model
1. User http://scrp.at/bmD
2. Group http://scrp.at/bmE
3. User_Groups http://scrp.at/bmF
4. User_Metadata http://scrp.at/bmG

and my Controller

http://scrp.at/bmH

WanWizard:
As I wrote before

> If $row[$pk_c] is not set at that moment, that can only mean $pk_c is still null, which means it couldn't find a primary key. Which would point towards a definition error in your model.

Both your User_Groups and User_Metadata tables do not have a primary key. ORM requires every model to have one!

purwandi:
Thanks, this is my fault :)

PR #150, opened 2012-05-11
FIX typing observer. Accept string as a date.

artbek:
MODEL:

'updated_at' => array(
    'label' => 'Updated At',
    'data_type' => 'time_unix',
)
- TYPING OBSERVER

'Orm\Observer_Typing' => array('before_save', 'after_save', 'after_load'),

Fieldset gets int from the mysql db and displays as date in the form.
But saving doesn't work as currently type_time_encode accepts only Date object - not string from the form.

The patch simply tries to convert supplied $var (if not Date already) to a Date object.

### DIFF
--- a/classes/observer/typing.php
+++ b/classes/observer/typing.php
@@ -335,11 +335,17 @@ class Observer_Typing
 	 * @return  int|string
 	 * @throws  InvalidContentType
 	 */
-	public static function type_time_encode(\Fuel\Core\Date $var, array $settings)
+	public static function type_time_encode($var, array $settings)
 	{
+
+		// if needed, try to convert to \Fuel\Core\Date object
 		if ( ! $var instanceof \Fuel\Core\Date)
 		{
-			throw new InvalidContentType('Value must be an instance of the Date class.');
+			try {
+				$var = \Date::forge($var);
+			} catch (\Exception $e) {
+				throw new InvalidContentType('Conversion to Date object failed. ' . $e->getMessage());
+			}
 		}
 
 		if ($settings['data_type'] == 'time_mysql')


WanWizard:
@jschreuder, comments?

jschreuder:
This will never work: the `Date::forge()` method takes a UNIX timestamp and that will never be returned by a form. Thus this doesn't do what you advertise as the thing this should solve.

Also it doesn't  follow our coding standards.

artbek:
You're right. Here's why it worked for me: <code>UpdatedAt</code>. 
It appears that the order of observers matters.
The problem happens while processing <code>before_save</code> event.

<pre><code>protected static $_observers = array(
  'Orm\\Observer_UpdatedAt',
  'Orm\\Observer_Typing',
);
</code></pre>

...doesn't work. 
It doesn't because <code>UpdatedAt</code> observer always returns a timestamp and then <code>Typing</code> observer tries to convert timestamp into a timestamp.

<pre><code>protected static $_observers = array(
  'Orm\\Observer_Typing',
  'Orm\\Observer_UpdatedAt',
);
</code></pre>

...works.

Would this be considered a bug?

steadweb:
This issue still exists within Orm. The order of observers as @artbek points out fixes the issue in hand.

WanWizard:
You should not define a data_type on the 'updated_at' or 'created_at' columns of your model, Observer_Typing won't touch properties without a defined datatype.

steadweb:
What's the correct way of typing the updated_at and created_at fields then? Instead of me doing the following:

<pre code="php">
echo \Date::forge($model->created_at)->format("%d/%m/%Y");
</pre>

When ideally I want the model to do this for me:

<pre code="php">
echo $model->created_at; // typing should cast this to an instance of \Fuel\Core\Date
</pre>

PR #151, opened 2012-05-14
Added where/order_by support for Has Many Relationship - Related to Issue #4

chadhutchins:
I added code to respect the relationship condition definitions, both where and order_by, when related records are being retrieved for has-many relationships. This same idea probably needs to be applied across all relationship types.

I implemented the code according to how the functionality is explained in the documentation: http://docs.fuelphp.com/packages/orm/relations/intro.html#config

Example Usage:

``` php
    protected static $_has_many = array(
        'comments' => array(
            'key_from' => 'id',
            'model_to' => 'Model_Comment',
            'key_to' => 'parent_id',
            'cascade_save' => false,
            'cascade_delete' => false,
            'conditions' => array(
                'where' =>  array(
                    array('model_class','=','Model_Page'),
                ),
                'order_by' => array(
                    'created_at' => 'DESC'
                ),
            ),
        ),
    );
```

### DIFF
--- a/classes/hasmany.php
+++ b/classes/hasmany.php
@@ -49,10 +49,31 @@ class HasMany extends Relation
 			next($this->key_to);
 		}
 		
+		// check for where conditions
 		foreach (\Arr::get($this->conditions, 'where', array()) as $key => $condition)
 		{
 			!is_array($condition) and $condition = array($key, '=', $condition);
 			$query->where($condition);
+		}
+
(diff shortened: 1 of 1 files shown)


WanWizard:
Implemented: https://github.com/fuel/orm/commit/fc5c0b3ad6642af7cdad5da9edeba5a71e7582e5

PR #155, opened 2012-05-22
Model get_diff on singular relation

felixgilles:
Fixed bug when the relation's new value is null


WanWizard:
This is a bugfix, so can you do a pull request on 1.2/develop?

And testing with empty() is dangerous, because NULL, 0, and '0' are valid values, but will evaluate to empty(). You might want to double check that.

PR #156, opened 2012-05-22
Model : initialisation of _original_relations. 

felixgilles:
Case: relation multiple with value is empty array

If it does not, in get_diff, for a multiple relation, _original_relations is not set and triggers a warning in in_array

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -720,6 +720,7 @@ class Model implements \ArrayAccess, \Iterator
 		{
 			if (is_array($data))
 			{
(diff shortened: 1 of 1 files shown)


WanWizard:
I don't think the idea here is to remove relations already present when a relations update is needed. That would reset already loaded relations.

@jschreuder ?

felixgilles:
I've don't master the ins and outs of what may already contain _original_relations. At worst, I could add an if !is_array() to the pull request.

PR #159, opened 2012-05-25
Model get_diff on singular relation

felixgilles:
Fixed bug when the relation's new value is null

I use empty() because I was inspired by the function code is_changed(). But I can use isset() if it is more appropriate (and also change is_changed())

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1322,11 +1322,13 @@ class Model implements \ArrayAccess, \Iterator
 			$rel = static::relations($key);
 			if ($rel->singular)
 			{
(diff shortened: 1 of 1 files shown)


WanWizard:
If empty() is used, it won't see the difference between not assigned, null, 0 and ""? Which makes the diff incomplete?

For is_changed() you have to be a bit more careful, because "0" and 0 doesn't mean it's changed (this happens a lot since the database driver retrieves numeric values as strings).

WanWizard:
I'll merge this manually (this pull request is on the wrong branch), I will deal with the empty() myself.

Issue #160, opened 2012-05-29
Relation 'conditions' option not taken into account when lazy-loading


Altreus:
If you want your relation's order_by to be taken into account, you have to eager-load it. This seems because if you access the relation and it hasn't been loaded already, the 'conditions' key in the relation's options isn't applied to the select query.

These should be taken into account when lazy-loading because to eager-load the relation when you forge the model would require the part of the code calling forge() to know that the relation needs to be eager-loaded. This seems to break encapsulation.

Altreus:
https://github.com/fuel/orm/pull/162

anthonycl:
This fix is on the right path but I believe it does not allow you to reference fields as where compares dynamically. I made some small modifications and added a fourth value to the where conditions array which if set to TRUE will use the compare value as the field from the originating model.

Example Belongs To Condition:
http://pastebin.com/tpJcmGKe

Replace line 59 on Altreus's modified belongsto.php:
$query->where($where);

With:
http://pastebin.com/uSsY7uqL

Again this was only temporary to get me by for now without spending too much time on this.

Thank You!

WanWizard:
@jschreuder when you have the time, this is a bug that's been there a while...

jschreuder:
Fixed by #162

PR #162, opened 2012-05-29
Use where and order_by when lazy loading relations

Altreus:
fixes https://github.com/fuel/orm/issues/160


kusmierz:
I haven't noticed, there is #160 and #162, so I made pull request (#169). Could somebody pull it to the main repo?

PR #165, opened 2012-06-02
Fixed Model::find_all

gustavosf:
The Model wasn't recognizing the method `Model::find_all`, only `Model::find_all_by_field` and `Model::find_by_field`. I used a regular expression to match only these three calls, avoiding calls like `Model::find_anything` or `Model::find_all_world_responses_that_make_no_sense` that wasn't raising exception

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -514,10 +514,10 @@ class Model implements \ArrayAccess, \Iterator
 		}
 
 		// Otherwise, lets find stuff
-		elseif (strpos($method, 'find_') === 0)
+		elseif (preg_match('/^find_(?P<type>all|((all_by_|by_)(?P<fields>[a-z_]+)))$/', $method, $match) === 1)
 		{
-			$find_type = strncmp($method, 'find_all_by_', 12) === 0 ? 'all' : (strncmp($method, 'find_by_', 8) === 0 ? 'first' : false);
-			$fields = $find_type === 'first' ? substr($method, 8) : substr($method, 12);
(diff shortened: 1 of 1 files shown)


WanWizard:
There was never any support intended for `Model::find_all()`. 
The synax to get all records is `Model::find('all')`.

So this fixes an issue that isn't there?

WanWizard:
@jschreuder, your opinion?

jschreuder:
As a preface: The callStatic usage was never really my thing to begin with, @philsturgeon added it for a nicer syntax but I always preferred the method chaining to this.

This pull-request looks logical to me, it was indeed never considered judging by the current code to be a necessary addition. But the code in here looks like a clean insertion that will not complicate the current code further or has the potential to break anything. As such I have no problem with it, just won't have the time to test it till late next week.

gustavosf:
@WanWizard you're right. Anyone who wants this feature can extend Model and include this kind of code by himself.

I just "grepped" documentation and found that [Model_Crud](http://docs.fuelphp.com/classes/model_crud/methods.html) already does that, and implements find_all with different arguments (it implements find_by & find_one_by too), but it's a core class, not orm.

So, it's not the case to mimic these magic methods behaviors between those classes? Or remove that from Orm/Model (looks like it is not documented yet)?

WanWizard:
@jschreuder didn't have any objections, so if you can resumit this pull request against 1.3/develop (the current develop branch), and provide a pull request for the modified documentation, I'll merge it.

jschreuder:
It's been a week, I'll close this over no interest. If any pull-request is made that does this I'll consider it anew.

PR #167, opened 2012-06-04
Adds support for intercepting object properties with getter/setter methods.

jellybeansoup:
Implement a method with the property name (i.e. property "id" becomes
method "id") to create a getter, and a method with the property name
prefixed by "set_" (i.e. property "id" becomes "set_id") to create a
setter.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -642,25 +642,31 @@ class Model implements \ArrayAccess, \Iterator
 			$new = false;
 		}
 
-		if ($new)
+		if( $new )
 		{
 			$properties = $this->properties();
 			foreach ($properties as $prop => $settings)
 			{
 				if (array_key_exists($prop, $data))
 				{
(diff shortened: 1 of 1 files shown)


jschreuder:
I'm sorry to be direct, but the quality of this pull-request is just too poor:
- It doesn't adhere to our [coding standards](http://docs.fuelphp.com/general/coding_standards.html)
- It contains commented out code
- Complicates `__set()` and `__get()` which should just forward to get() & set() to keep the generic getter()/setter() logic in one place

And besides that the code calls a magic method directly (which is strangely resource expensive in PHP) and does that for each and every call in the constructor. If you are loading a lot of objects this will slow down the hydration to such a massive amount that it'll be a huge drag on the system.

The bottom line is that this is something you like that will slow down the system for everyone, a tradeoff that is not worth making.

PR #169, opened 2012-06-11
Uses "conditions" settings even when not in "related"

kusmierz:
So now, if I'm not using "related", I can sort result.


jschreuder:
Fixed by #162

kusmierz:
@jschreuder Thanks

PR #186, opened 2012-10-02
is_changed('field') was always returning true

lukearmstrong:
Broken here, by a seemingly unrelated commit.

https://github.com/fuel/orm/commit/df62d1115888a81bff822ae9a038d823b9a3fbc2

`is_changed('field')` will always return `true`.

For now I have just removed the code that was added, as it isn't clear what you were trying to achieve.


lukearmstrong:
Fixed by https://github.com/fuel/orm/commit/05e73449042935b65b3705c7e41fdc4fc3f0440e

PR #188, opened 2012-11-12
Soft delete ORM model.

emlynwest:
Implemented a soft delete ORM model.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -18,6 +18,9 @@ Autoloader::add_classes(array(
 	'Orm\\HasOne'       => __DIR__.'/classes/hasone.php',
 	'Orm\\ManyMany'     => __DIR__.'/classes/manymany.php',
 	'Orm\\Relation'     => __DIR__.'/classes/relation.php',
+	
+	//Speclised models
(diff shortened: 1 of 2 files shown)


WanWizard:
You should add "RelationNotSoft" to the package bootstrap to avoid load issues.

Also run a quick check on coding standards, for example

``` php
// if (!something) ...should be if ( ! something)...
// $deletedColumn should be $deleted_column
// do you really mean to use self:: instead of static:: ?
```

And I think you missed Model::query(), which needs to have the filter included too.

emlynwest:
Thanks for the comments Wan. I'll go through and make the changes.

Also the reason I was using `self` is that using `static` produces a `Cannot access property` exception.

WanWizard:
Perhaps you can have a look at whether the __callStatic is absolutely needed for that delete method.

emlynwest:
It's not needed to call the delete method. It's added as a convenience (see my reply on https://github.com/stevewest/orm/commit/f90d051f29308e6d6a30a334c4fd68988a969cb3#classes-model-soft-php-P75 )

PR #193, opened 2012-12-20
clear cached object(s) for memory leak

hackoh:
I wonder memory leaking

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -199,6 +199,24 @@ class Model implements \ArrayAccess, \Iterator
 
 		return $result;
 	}
+	
+	/**
+	 * Clear cached object(s)
+	 * 
+	 * @param void
+	 */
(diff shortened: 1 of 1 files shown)


hackoh:
sorry, i mistaked newline character...

WanWizard:
Before you open it again, can you explain why you think there's a memory leak, and show us how you have determined there is one?

hackoh:
When insert a record, the instance is always contains in $_cached_objects.
(classes/model.php:L1182)
For example, if I implement the "csv import" process, this will put pressure on memory.
Actually, I want to setup contains or not in config, But first, I want a way to explicitly remove.

(I reopened #194)

WanWizard:
That is not a memory leak, that is by design. The ORM caches are records it retrieves.

You should not use the ORM for batch operations, it's not what it is designed for.

PR #194, opened 2012-12-20
Clear cached object(s)

hackoh:
This commit is a solution for the memory leak

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -199,6 +199,24 @@ class Model implements \ArrayAccess, \Iterator
 
 		return $result;
 	}
+	
+	/**
+	 * Clear cached object(s)
+	 *
+	 * @return  void
+	 */
+	public static function clear_cached_object($id = null, $class = null)
+	{
+		$class = $class ?: get_called_class();
+		if ($id === null)
+		{
+			unset(static::$_cached_objects[$class]);
+		}
+		else
+		{
+			unset(static::$_cached_objects[$class][$id]);
+		}
+	}
 
 	/**
 	 * Get the primary key(s) of this class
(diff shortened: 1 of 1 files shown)


frankdejonge:
@WanWizard will be the judge of wether this'll be merged. I'll add some code comments in case it does.

classes/model.php
```diff
@@ -199,6 +199,24 @@ public static function cached_object($obj, $class = null)
 
 		return $result;
 	}
+	
+	/**
+	 * Clear cached object(s)
+	 *
```
frankdejonge, line 205: Docsblok is missing the `@param`'s

classes/model.php
```diff
@@ -199,6 +199,24 @@ public static function cached_object($obj, $class = null)
...
+	 * @return  void
+	 */
+	public static function clear_cached_object($id = null, $class = null)
+	{
+		$class = $class ?: get_called_class();
```
frankdejonge: $class should only be `get_called_class()`. One model can never influence the caching of another. The class param should be removed from the arguments list.

classes/model.php
```diff
@@ -199,6 +199,24 @@ public static function cached_object($obj, $class = null)
...
+		if ($id === null)
+		{
+			unset(static::$_cached_objects[$class]);
```
frankdejonge: I would use the Arr class to remove it, it might not be there and therefor result in a php error/notice.

classes/model.php
```diff
@@ -199,6 +199,24 @@ public static function cached_object($obj, $class = null)
...
+		}
+		else
+		{
+			unset(static::$_cached_objects[$class][$id]);
```
frankdejonge: Same here.

WanWizard:
See #193. This is not a memory leak, this is incorrect use.

PR #202, opened 2013-01-10
Temporal/Revisions Model

emlynwest:
Implements temporal/revisions for ORM models.

Sorry for the 3rd pull request. I am new at this and finally managed to work out how to get git to squash :) Sorry for all the alert emails.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -20,7 +20,9 @@ Autoloader::add_classes(array(
 	'Orm\\Relation'     => __DIR__.'/classes/relation.php',
 	
 	//Speclised models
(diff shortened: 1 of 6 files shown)


WanWizard:
See inline comments. I see also code that allows PK checking to be enabled/disabled, but I don't see it being used anywhere?

emlynwest:
The PK checking is used 6 times currently, https://github.com/stevewest/orm/commit/f4be5d6ac6ac1b1b4ab422384e1e9b13695f8942#classes-model-temporal-php-P312 for example.

I did go through and check for cammelcase, I must have missed one and I'll go through again to check comments and the like.

Issue #171, opened 2012-06-13
Relations are not fetched for new objects


Savageman:
It's basically all explained in this ScrapYrd: http://scrp.at/buD#L47

A new item won't try to fetch a relation. I'm wondering why it is that way?

In the example, I create a new object, set a property so the relation can properly be found if I request for it. But it's not fetch since it's a new item. Seems a bit strange to me.
I'm not sure whether it's intended or not, that's why I'm asking!

Thanks.

WanWizard:
A new item can never have a relation, you can't relate to something that does not exist yet.

It would be possible to fetch them for relations where you have the FK, but not in all other cases as a new object doesn't have a primary key yet.

Savageman:
A `belongs_to` relationship needs only a FK to exists, the PK is not necessary. Maybe I didn't insist much on this in my example. `belongs_to` feels like a special case maybe. It should at least try to get the remote model.
I agree nothing can be fetched for other relation types though (as they a need a PK to fetch the data).

I was thinking the logic of checking whether it's a new object or not should be in the `get()` methods within the relations, as the need of the PK depends on the type of relation we have?

WanWizard:
That's what I wrote, if you have a valid FK then you should be able to fetch them.

The relationship definition should tell you if the FK is in the current or in the other model.

PR #205, opened 2013-01-14
Update classes/model.php

peter-fritchley:
fixes a return of the issue in #148

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -989,8 +989,15 @@ class Model implements \ArrayAccess, \Iterator
 		{
 			if ( ! array_key_exists($property, $this->_data_relations))
 			{
-				$this->_data_relations[$property] = $rel->get($this);
-				$this->_update_original_relations(array($property));
+				if ($this->is_new())
+				{
+					$this->_data_relations[$property] = $rel->singular ? null : array();
+				}
+				else
+				{
+					$this->_data_relations[$property] = $rel->get($this);
+					$this->_update_original_relations(array($property));
+				}
 			}
 			return $this->_data_relations[$property];
 		}


WanWizard:
Any idea what caused the issue to reappear? 

I can't find any recent changes to this code, and it must come from somewhere. I want to avoid fixing something here that is caused elsewhere...

peter-fritchley:
hmm.. seems that the change was reverted in response to https://github.com/fuel/orm/issues/171 - although in this case the fk is not set.

This causes problems because breaking the relation between models is implemented as setting the foreign key to null - meaning that these will be fetched by accessing the relation property for any newly created model.

WanWizard:
From what I can see, and reading back the comments in the first issue, the point is that it needs a check if the FK is present before trying to fetch it? So it only fetches related objects when the FK !== null?

peter-fritchley:
Ok, well in my case the relation is being fetched even though the FK === null.

Do you think it is the responsibility of the relation class to determine whether the FK is acceptable?

WanWizard:
That's what I mean. The fix to this issue should be "do not fetch if FK === null".

The original code (and the one you propose the put back in) focuses on the state of the object (whether or not it has a PK) while it should check if there is an FK available for the relation being accessed, and only try to get it if there is one.

peter-fritchley:
Ok, I'll add another pull request with it implemented within the relation classes.

Thanks.

WanWizard:
No worries, I'm already on it. Just wanted to check if I was on the right track ;)

WanWizard:
Just double checking: this is only an issue on a "belongs_to", right? It's the only relation where the model itself has the FK.

PR #206, opened 2013-01-15
Update classes/manymany.php

peter-fritchley:
ManyMany relations should return array when FK has not been set as they are not singular.


WanWizard:
Chips, you're right. sorry 'bout that.

Issue #213, opened 2013-01-28
Recent commit broke things


phil-lavin:
Commit 2c369de3ee19a2700c3fafb391eec2f040ba62b5

With this ORM query...

``` php
$menu = \Menus\Model_Menu::find()
         ->where('title', 'Food')
         ->related('categories')
         ->where('categories.live', true)
         ->order_by('categories.position')
         ->get_one();
```

The SQL goes from this...

``` sql
SELECT `t0`.`id` AS `t0_c0`, `t0`.`title` AS `t0_c1`, `t0`.`position` AS `t0_c2`, `t0`.`live` AS `t0_c3`, `t0`.`pp_enabled` AS `t0_c4`, `t0`.`pp_image` AS `t0_c5`, `t0`.`created_at` AS `t0_c6`, `t0`.`updated_at` AS `t0_c7`, `t1_through`.`category_id`, `t1_through`.`menu_id`, `t1`.`id` AS `t1_c0`, `t1`.`title` AS `t1_c1`, `t1`.`alias` AS `t1_c2`, `t1`.`type` AS `t1_c3`, `t1`.`position` AS `t1_c4`, `t1`.`live` AS `t1_c5`, `t1`.`visible` AS `t1_c6`, `t1`.`button` AS `t1_c7`, `t1`.`banner` AS `t1_c8`, `t1`.`banner_alt` AS `t1_c9`, `t1`.`banner_link` AS `t1_c10`, `t1`.`page_title` AS `t1_c11`, `t1`.`page_description` AS `t1_c12`, `t1`.`page_keywords` AS `t1_c13`, `t1`.`created_at` AS `t1_c14`, `t1`.`updated_at` AS `t1_c15` FROM `menus__menus` AS `t0` LEFT JOIN `menus__menus_categories` AS `t1_through` ON (`t0`.`id` = `t1_through`.`menu_id`) LEFT JOIN `menus__categories` AS `t1` ON (`t1_through`.`category_id` = `t1`.`id`) WHERE `t0`.`title` = 'Food' AND `t1`.`live` = '1' ORDER BY `t1`.`position` ASC
```

To this...

``` sql
[...]

PR #215, opened 2013-01-29
Update classes/query.php

peter-fritchley:
return the default join result in modify_join_result, fixes issues experienced in #213


WanWizard:
Already thought this was the issue. /cc @stevewest

emlynwest:
Was waiting for people posting the original issue to get back to see if that was indeed the problem. Glad that it's sorted though.

Issue #220, opened 2013-02-20
Using Model_Temporal::find() then lazy loading temporal related models returns all versions of related


iturgeon:
Using find() on the parent and lazy loading the related model will return all revisions.

``` php

<?
class Model_Vehicle extends Orm\Model_Temporal
{
    protected static $_properties = [
        'id',
        'name',
        'temporal_start',
        'temporal_end',
    ];

    protected static $_has_many = ['riders'];
    protected static $_temporal = ['mysql_timestamp' => false];
}

class Model_Rider extends Orm\Model_Temporal
{
    protected static $_properties = [
        'id',
        'name',
        'vehicle_id',
        'temporal_start',
        'temporal_end',
    ];

    protected static $_temporal = ['mysql_timestamp' => false];
}
```

``` php
$car = Model_Vehicle::forge(['name'=>'ford']);
$dog =  Model_Rider::forge(['name' => 'Sparky']);
$car->riders[] = $dog;
$car->save();

$dog->name = 'Turbo';
$dog->save();

$dog->name = 'Ruffio';
$dog->save();

$car->name = 'Chevy';
$car->save();

$chevy = Model_Vehicle::find(1);
$chevy->riders // returns an array of all the riders AND all rider revisions related to this vehicle
```

This also affects `find_revision::(1, null)` as it internally runs `find()`

emlynwest:
Is this still a problem after #221?

emlynwest:
Problem seems to have come back for both eager and lazy loaded relations. Not quite sure what's caused it but looking into it now.

emlynwest:
Closed with d18ec214e4ac027da12d602dbd68ce273c337dc1

PR #221, opened 2013-02-20
Fixed lazy loading of temporal related models when using find() #220

iturgeon:
This also solves a problem when get() didn't directly return the results of parent::get() causing and `Indirect modification of overloaded property` error


WanWizard:
@stevewest, when you have time?

emlynwest:
Looks good to me.

PR #224, opened 2013-02-21
Fixed bug about obtaining args in Model_Soft::__callStatic

pirosikick:
The following code fails because using `array_pop` to obtain arg. I think using `array_shift` is correct in this case.

``` php
class Model_Something extends Model_Soft { }

// error
Model_Soft::find_deleted('all', array(...));
```

Signed-off-by: Hiroyuki Anai pirosikick@gmail.com

### DIFF
--- a/classes/model/soft.php
+++ b/classes/model/soft.php
@@ -128,8 +128,8 @@ class Model_Soft extends Model
 		{
 			$temp_args = $args;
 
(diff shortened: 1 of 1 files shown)


WanWizard:
@stevewest when you have time?

emlynwest:
Not sure why this did not come up earlier.
@WanWizard Looks good to merge in.
@pirosikick Thanks for the fix.

PR #226, opened 2013-02-25
Add the possibility to use Fuel\Core\Cache in order to cache model properties.

felixgilles:
Cache is automatically updated when accessing a non-existent cached property.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -292,7 +292,7 @@ class Model implements \ArrayAccess, \Iterator
 		{
 			try
 			{
-				$properties = \Cache::get('model_properties.'.$class);
+				$properties = \Cache::get('model_properties.'.str_replace('\\', '_', $class));
 			}
 			catch (\CacheNotFoundException $e) {}
 		}
@@ -303,7 +303,7 @@ class Model implements \ArrayAccess, \Iterator
 			try
 			{
 				$properties = \DB::list_columns(static::table(), null, static::connection());
-				$use_cache_file and \Cache::set('model_properties.'.$class, $properties);
+				$use_cache_file and \Cache::set('model_properties.'.str_replace('\\', '_', $class), $properties);
 			}
 			catch (\Exception $e)
 			{
@@ -1042,7 +1042,7 @@ class Model implements \ArrayAccess, \Iterator
 			if ($reload_cache)
 			{
 				$class = get_called_class();
-				\Cache::delete('model_properties.'.$class);
+				\Cache::delete('model_properties.'.str_replace('\\', '_', $class));
 				unset(static::$_properties_cached[$class]);
 				$reload_cache = false;
 				try
--- a/classes/model.php
+++ b/classes/model.php
@@ -117,6 +117,11 @@ class Model implements \ArrayAccess, \Iterator
 		'many_many'     => 'Orm\\ManyMany',
 	);
 
+	public static function _init()
+	{
+		\Config::load('orm', true);
+	}
+
 	public static function forge($data = array(), $new = true, $view = null)
 	{
 		return new static($data, $new, $view);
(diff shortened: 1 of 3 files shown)


emlynwest:
I see no actual changes to temporal and it looks like you are using an old version of the orm, there have been several fixes recently so it might be a good idea to merge the changes in and check that everything is ok.

felixgilles:
Wait, I identified an issue : I tested with a class without namespace. Within a namespace, the cache identifier contains backslash and Cache throws an Exception.

felixgilles:
@stevewest I will check for the version. I made one change in temporal, I moved the _init() method intto Model class.

emlynwest:
@felixgilles I did not see that at first, my bad

WanWizard:
I don't think this is a good idea. 

The code to dynamically fetch table columns and create properties on the fly was put in to make development easier, but once it's finished you should define the properties in the model. You will have to anyway if you want to do something useful with it, like fieldset or validation.

I don't think caching it to work around not having to define your properties is a good idea or best practice.

felixgilles:
Hi Harro,

Your arguments have convinced me personally to declare properties. But the subject has generated a debate in Novius (we use FuelPHP for Novius OS). To summarize, our general opinion is: if there is magic, it should be everywhere. If a developer codes without declaring Model properties, no incentive exists to change its current approach. When publishing a website to production, developpers fear to change anything because of bugs that could be added during the process.

This PR allows these people to continue to code their own way (certainly not good practice but it works) but optimizes the system. And other users won't be impacted (it just loads an additionnal config file)

However, I think some changes should be done in order to stimulate properties declaration:
- the Orm should automatically (with "SHOW COLUMNS") update properties only if some settings are enabled.
- Change the documentation; the first model declaration should have properties defined, since it is best practice.

WanWizard:
Still a bad idea, and as you say, it promotes bad practice.

Another issue is that it's very MySQL centric, as it's the only driver that reliably produces column information. So if you rely on this feature in your code, and you (or the user of your product) decides to switch from MySQL to another RDBMS, they are in deep trouble...

I won't remove it in 1.x because of backward compatibility, but this discussion has lead me to think about deprecating it, and remove it for 2.0.

felixgilles:
Deprecating this feature seem to me a good idea if you want to promote good practices. If you do that, this PR has no longer interest. You can close it.

felixgilles:
You've not close this PR. Is it an omission or the decision is not yet definitive ?

After some research, I found that others ORM caches the tables schema too, for example: [Rails](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaCache.html) or [PHP ActiveRecord](http://www.phpactiverecord.org/docs/ActiveRecord/Config)

It seems to be a quite universal practice...

And, it's possible to retrieve columns description with the PDO driver for PostGreSQL : [Zend DB](https://trac.agron.iastate.edu/trac/nwnwebsite/browser/include/Zend/Db/Adapter/Pdo/Pgsql.php?rev=5131#L129) or [Doctrine](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php#L290). 
It's MySQL Centric only because the PDO driver does only a "DESCRIBE table", but this is fixable.

WanWizard:
I know some ORM's cache table schema's (my Datamapper for CI does that too), because it's the only way the model knows what the table's properties are. In Fuel's ORM, you have (err, should) define your properties in the model definition, so it doesn't need the schema. 

It is only there to "aid" development, so you don't have to update your models every time you make a schema change.

It also creates a dependency on the Cache class I'm not too happy about.

I've left this open because i understand why you would want this, but I don't like the implementation (and it's limitations).

felixgilles:
I understand your point of view for the Cache class dependency. For a standalone package, it's not a good idea.

I can try to modify the implementation and bypass this limitation. What do you suggest? Adding a cache mechanism in the ORM package?

felixgilles:
Maybe I could implement a Cache Interface for the ORM and use a Driver Pattern. The default config could use the FuelPHP Cache.

emlynwest:
Adding cache functionality to the orm is not good either, as it's an orm, not a cache. As @WanWizard said, the best practice is to fully define your model's properties.

WanWizard:
Thought long and hard about this one, and finally decided to veto it. 

Dynamically loaded properties is handy in development, but once you're done, define them in the model. In a production environment you don't want additional queries and/or cache hits anyway.

felixgilles:
Thank for your response, even if it's not positive !

For the future of our CMS application, can you tell me if you are going to deprecate this feature (fetching columns properties by DB query)? If yes, only in PRODUCTION or for all environments?

emlynwest:
I can't see the feature being deprecated.

WanWizard:
No, no plans. Its serves it's purpose, it's up to the developer that the code going into production doesn't rely on a DESCRIBE query every time a model loads...

PR #227, opened 2013-02-28
Added default primary keys to Model_Temporal

iturgeon:
If a Temporal Model does not set the primary keys some very strange behaviors can sneak up on you. 

Given the model has defaults for the names of these columns, I think it would be a good idea to give the model a matching default primary key.

### DIFF
--- a/classes/model/temporal.php
+++ b/classes/model/temporal.php
@@ -11,6 +11,11 @@ class Model_Temporal extends Model
 {
 
 	/**
+	 * Compound primary key that includes the start and end times is required
+	 */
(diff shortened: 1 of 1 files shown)


frankdejonge:
@stevewest?

emlynwest:
I will take a look later today. I did look into this originally but I think I came across a problem. I think I had a good reason for it but can't recall at the moment (so can't have been that good). The PR looks fine but I'd just like to double check.

emlynwest:
Looks fine to merge.

PR #230, opened 2013-03-01
Eager Loading of non-temporal related models from a temporal model breaks.

iturgeon:
The temporal Model's `modify_join_result()` assumes that all joins on related models are also temporal models, adding temporal query options to them.

This checks if the related model is temporal before inserting the temporal query options into the join.

### DIFF
--- a/classes/temporal/query.php
+++ b/classes/temporal/query.php
@@ -43,7 +43,7 @@ class Temporal_Query extends Query
 	 */
 	protected function modify_join_result($join_result, $name)
 	{	
-		if( ! is_null($this->timestamp) and is_a($join_result[$name]['model'], '\\Orm\\Model_Temporal'))
+		if( ! is_null($this->timestamp) and is_subclass_of($join_result[$name]['model'], '\Orm\Model_Temporal'))
 		{
 			//Add the needed conditions to allow for temporal-ness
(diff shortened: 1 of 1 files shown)


iturgeon:
Might as well turn on the bat signal again: @stevewest

emlynwest:
@iturgeon No need, saw the link from your comment in the issue you created.

WanWizard:
@stevewest just ping me if I need to merge something. ;)

emlynwest:
@WanWizard This looks good to merge. (Thank you!)

emlynwest:
Scratch that, using, `is_a` does not return the correct result when checking if the model is a subclass of another.

The behaviour I was seeing:

``` php
is_a('Orm\Model_Temporal', 'Model_MyTemporal') === false;
is_subclass_of('Model_MyTemporal', 'Orm\Model_Temporal') === true;
```

http://www.php.net/manual/en/function.is-a.php#107181

iturgeon:
Ah right, I got the same results from your tests on `is_a`.

Updated to `is_subclass_of` and used the same model name string format as is used in hydrate (single back slashes).

emlynwest:
@WanWizard This is good to merge in now :)
@iturgeon Thanks!

WanWizard:
@stevewest Yes sir... :smiley:

PR #231, opened 2013-03-04
Phpdoc improvments on Query class

SCIF:
I rewrite some definitions and typed returning of min/max methods. Some other cosmetics. I have in plans same actions in model class. What do you think?

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -20,11 +20,11 @@ class Query
 	/**
 	 * Create a new instance of the Query class.
 	 *
-	 * @param	string	name of the model this instance has to operate on
-	 * @param	mixed	DB connection to use to run the query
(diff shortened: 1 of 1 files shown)


classes/query.php
```diff
@@ -1266,13 +1278,13 @@ public function min($column)
 			return false;
 		}
 
-		return $min;
+		return (int) $min;
```
frankdejonge, line 1281: @WanWizard this is a BC breaking change in theory but does seem to be better than it was before. Agreed? btw, awesome work @SCIF :+1:

WanWizard:
@FrenkyNet I can live with it, I can't imagine someone explicitly wanting a string returned from a min() or max() method.

AddoSolutions:
Nice job on this :+1: 

Thanks!

PR #236, opened 2013-03-09
Propose set() extension about has_many

hackoh:
Please see https://github.com/hackoh/orm/blob/1.6/model_tests/tests/model.php

This change makes set() method available to set  to a has_many-related property using assoc. 

For example, can receive input names like `comments[][title]`(inserting), `comments[10][title]`(updating)  from the form, the controller can set() simply.

What do you say to this extension?

(I don't know how to implement tests in case of Orm, so this branch does not include that..sorry.)

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1068,7 +1068,7 @@ class Model implements \ArrayAccess, \Iterator
 			{
 				$this->_data[$property] = $value;
 			}
-			elseif ($relation = static::relations($property))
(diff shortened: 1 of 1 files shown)


WanWizard:
For loading or updating the object from a posted form, especially one created with a fieldset and/or a tabular form, you can use from_array(). It's a lot more efficient then using set() calls.

I've added the logic to it to support creation of new related objects to it.

PR #239, opened 2013-03-14
Moves config load call to be more reliable

emlynwest:


WanWizard:
Did you encounter a situation where _init() isn't called?

emlynwest:
I did, yes. I was not able to find out why it was not being called and thought it safer to load the config only where it's actually used.

frankdejonge:
@stevewest better would be to find out why the init function isn't called and fix that, this shouldn't happen. Also this will decrece performance.

emlynwest:
@FrenkyNet It would but I did look into it and was not able to find out why. I'll take another look later today though.

frankdejonge:
I'm closing this one as this isn't the fix...

PR #245, opened 2013-03-29
Allow ordering on through table in many-many relationships

rob-mccann:
Previously one could only order on the `model_to`'s table.

Now, you can order on the `through` table too which allows for instances like below where the user wants to change the order that associated images are in.

This'll all need properly checking over for regressions by someone that knows more about the ORM's inner-workings than I.

``` php
class Model_Post extends \Orm\Model
{
protected static $_many_many = array(
    'images' => array(
        'conditions' => array(
            'order_by' => array(
                'images_posts.position' => 'asc',
            )
        )
    )
);

}
```

### DIFF
--- a/classes/manymany.php
+++ b/classes/manymany.php
@@ -151,11 +151,14 @@ class ManyMany extends Relation
 	{
 		$alias_to = 't'.$alias_to_nr;
 
+		$alias_through = array($this->table_through, $alias_to.'_through');
+		$alias_to_table = array(call_user_func(array($this->model_to, 'table')), $alias_to);
+
 		$models = array(
 			$rel_name.'_through' => array(
 				'model'        => null,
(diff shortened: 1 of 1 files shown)


WanWizard:
I have  mixed feelings about this one.

A through table should not contain anything but the two FK's. If it contains attributes too, it should have a Model associated with it, converting the many-2-many into two has-many's with the new model in the middle.

emlynwest:
I think this would be a useful thing (not looked at the code yet). I have needed to be able to order/access the through table before when applying settings to the relation in this way.

rob-mccann:
@wanwizard definitely see where you're coming from. It just seemed a bit like overkill to have a whole object just so I could order the relation :)

emlynwest:
@rob-mccann If you have not already (not checked), could you create some docs for this too please?

WanWizard:
Already done that. ;)

PR #248, opened 2013-04-03
Modified __isset() to handle `false` attribute values

ghost:
When checking `isset()` on an EAV attribute, if the value of the attribute was false then `isset()` would return false. I don't think this the correct outcome.

``` php

// An example
// assuming that $product is an orm model with an 'attributes' relation for EAV data
$product->attributes[] = \Model_Attribute::forge(array('key' => 'any_key', 'value' => false));
$product->save();

isset($product->any_key); // would return false, should return true

```

Previously if you wanted to check whether a attribute key was set you couldn't. I've made some changes to alleviate this, if an attribute key exists then `isset()` should return true, otherwise false.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1746,8 +1746,7 @@ class Model implements \ArrayAccess, \Iterator
 	 * 
 	 * @param string $attribute, the attribute key to check
 	 *
-	 * @return void
-	 * @author James Pudney james@phpgenie.co.uk
+	 * @return boolean
 	 **/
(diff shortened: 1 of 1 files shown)


WanWizard:
It was implemented this way in the assumption that it is pointless to store booleans because you can't store them anyway, but I guess some RDBMSs can.

I find that this proposed fix duplicates a lot of code, so I took the liberty of implementing it in a different way.

PR #251, opened 2013-04-05
get_class($this) in case of extended class

vrcAlbert:
As I can see in the file, it's been done for other methods, it could be usefull to also do this here.


vrcAlbert:
Thanks for dealing with this PR, I'm sorry I did not had the time to amend it by myself...

PR #262, opened 2013-05-21
Fixed notice and phpdoc improvements

SCIF:
Phpdoc fixes for Model and Relarion classes.

Commit 5ef1e6884b01b067c81dcc3d78e3a2e460545617 must be ported to 1.6/master — bug is there also.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1628,6 +1628,7 @@ class Model implements \ArrayAccess, \Iterator
 			$rel = static::relations($key);
 			if ($rel->singular)
 			{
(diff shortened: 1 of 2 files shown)


WanWizard:
There will be a 1.6.1. soon, for which we're going to backport all bugfixes.

SCIF:
Fix is still not ported.

https://github.com/fuel/orm/blob/1.6/master/classes/model.php#L1574

WanWizard:
I am very aware of that, trying to find time for it. Perhaps this weekend...

PR #270, opened 2013-06-14
Added Collections to Orm Data Results

AddoSolutions:
This change is a proof of concept of extendable per-model data collections being set up for any ORM dataset, while still not interfering with existing code.

In short, instead of returning a static array, this would return an object with ArrayAccess enabled on it so that you can still treat it like an array, but each model can custom define a collections model so that one has infinite flexibility on the data that is returned.

I am confident there is a more efficient way of generating these collections, however this will get the ball rolling.

Feel free to contact me with further questions on this.

Thanks,

### DIFF
--- a/classes/collection.php
+++ b/classes/collection.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * Fuel
+ *
+ * Fuel is a fast, lightweight, community driven PHP5 framework.
+ *
+ * @package    Fuel
+ * @version    1.6
+ * @author     Fuel Development Team
+ * @license    MIT License
+ * @copyright  2010 - 2013 Fuel Development Team
+ * @link       http://fuelphp.com
+ */
+
+namespace Orm;
+
+class Collection extends \ArrayIterator
+{
+	private $_model;
+
+	/**
+	 * forge function.
+	 * 
+	 * @access public
+	 * @static
+	 * @param mixed $model
+	 * @param array $data (default: array())
+	 * @return void
+	 */
+	public static function forge($model,array $data = array()){
+		return new self($model, $data);
+	}
+	
+	/**
+	 * __construct function.
+	 * 
+	 * @access public
+	 * @param array $array (default: array())
+	 * @return void
+	 */
+	public function __construct($model, array $array=array()) {
+		$this->_model = $model;
+		
+		return parent::__construct($array);
+	}
+	
+	/**
+	 * getMode function.
+	 * 
+	 * @access public
+	 * @return void
+	 */
+	public function getMode(){
+		return $this->_model;
+	}
+	
+	public function implode_pk($data){
+		$data = array();
+		
+		foreach($data as $n=>$v){
+			$data[] = $v->implode_pk($v);
+		}
+		
+		return $data;
+	}
+}
--- a/classes/collection.php
+++ b/classes/collection.php
@@ -44,13 +44,47 @@ class Collection extends \ArrayIterator
 		return parent::__construct($array);
 	}
 	
+	
+	/**
+	 * __get function.
+	 * 
+	 * @access public
+	 * @param mixed $name
+	 * @return void
+	 */
+	public function __get($name){
+		
+		$class = get_called_class();
+		$results = new $class($this->get_model());
+		
+		$is_nested = true;
+		
+		foreach ($this as $n=>$v){
+			$results[$n] = $v->{$name};
+			
+			if (!is_subclass_of($v,$class) && get_class($v) != $class) $is_nested = false;
+			
+		}
+		
+		if (count($this) && $is_nested){
+			$_results = array();
+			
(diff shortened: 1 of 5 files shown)


classes/collection.php
```diff
@@ -0,0 +1,66 @@
...
+		return parent::__construct($array);
+	}
+	
+	/**
+	 * getMode function.
+	 * 
+	 * @access public
+	 * @return void
+	 */
+	public function getMode(){
```
frankdejonge: this should be named `get_mode`

classes/collection.php
```diff
@@ -0,0 +1,66 @@
...
+		return $this->_model;
+	}
+	
+	public function implode_pk($data){
```
frankdejonge, line 98: the docblock is missing

classes/collection.php
```diff
@@ -0,0 +1,66 @@
...
+		$data = array();
+		
+		foreach($data as $n=>$v){
```
frankdejonge: the bracket should be on a new line

frankdejonge:
If this were to be included I'd suggest having the default return value to be plain array. When the collection setting is set to true, use the Orm\Collection and when string a new instance of that class. @WanWizard thoughts?

classes/collection.php
```diff
@@ -0,0 +1,66 @@
...
+	 * forge function.
+	 * 
+	 * @access public
+	 * @static
+	 * @param mixed $model
+	 * @param array $data (default: array())
+	 * @return void
+	 */
+	public static function forge($model,array $data = array()){
+		return new self($model, $data);
```
frankdejonge: this should be `new static`

emlynwest:
With only a quick look at the code I would agree with @FrenkyNet in making the default an array, there could be repercussions regarding BC if an array is suddenly no longer returned.

AddoSolutions:
Interesting, thanks for all the feedback!

I will get those changes made and recommit.

I also made an update to the code that had caused an issue with making model-level collections.

Once accepted, should I swing over to fuel/docs and make an update there to the orm?

emlynwest:
Docs updates are always good! Especially for new features.

WanWizard:
This is an interesting idea, but I don't like the implementation.

This way it's much to intrusive, it does config loads on every get, it uses a config value that isn't needed (the model can tell you if a collection class needs to be used or not, and the code is far from Fuel standard.

It would be better if the collection model is passed as part of the query options array. The model is aware of it, so it can pass it. And then Query can use it (or not) without further interaction.

If it is the idea that you can have custom collection implementations, then probably this should be an interface, and not a class. As is it doesn't give you any benefit, so why not use array() by default, and create custom collection classes implementing the interface if needed?

AddoSolutions:
@WanWizard In regards to the implementation, I agree fully, this was a proof oc concept to get the reaction of the community.  I want to make sure it is worth my time to go through and refine it to a production level quality.

In regards to your comment about it not giving any benefit, I would agree that at it's current phase, it does not, however as the ORM develops further, I am confident that various tools will be added to this class (such as getIds(), perhaps a save() function, a priority queue, etc. could be added as time goes on.

More importantly, if people begin developing with collections, it then alleviates the need to possibly re-do code (as minimal as it will be), and can delve right into developing the respective collections classes.

Once the implementation is cleaned up and furthermore more efficient, I think this would be an excellent and un-intrusive addition that would greatly benefit a great deal of people, and those that chose not to use it would be completely unaffected.  I know that last several applications that I used FuelPHP for, collections would have saved me a great deal of inefficient code.

Let me know what you guys think, and if you want to proceed, I will go ahead and clean this up so it is merge-ready.

AddoSolutions:
I knew I was forgetting something!

Check out that commit, that was my original inspiration to build this in the first place.

Again, it isn't fully optimized for performance, but it a proof of concept and great demonstration of this.

Here is an example of the power of this code:

You have three models:

```
php oil g scaffold person name:string email:string
php oil g scaffold post name:string content:string
php oil g scaffold comment post_id:int person_id:int content:string
```

Lets assume the proper relations are in place (People have many posts – People have many comments). So we have out Mosel_Post loaded into $post.  Watch this:

```
$post->comments->people->name
```

Returns a collection of all of the names of the people that have commented on that post and furthermore:

```
implode(", ", (array)$post->comments->people->name);
```

And boom, you have a comma separated list of these people in one line of code!

I suspect the same would be true with methods, but that could get a little dicier.

Anyways, I think that could be a game-changer for accessing data in the orm.

Let me know what you guys think.

WanWizard:
I absolutely love the idea, reminds me of my CI ORM. ;)

The implementation is still not up to standards though, and this PR is slowly becoming a mess with all the different changes, it can't be be merged like this.

I suggest you finish it, make sure it complies to the standards, rebase your branch, and then send in a new PR.

WanWizard:
Will be addressed in the 2.0 ORM.

PR #271, opened 2013-06-15
Active Model Record Implementation

AddoSolutions:
This is a pretty simple concept, in most occasions, there is a particular record that defines the current request (whether that is the currently logged-in user, the current post you are looking at, or the current category, etc.).

Basically this gives the ability to load once that active record, cache it, and easily and quickly load it in the the future.

Example:

```
Model_Post::find(1234)->set_active_record();
Model_User::find(321)->set_active_record();
```

That data is then stored the the model class and is then can be extremely easily recalled like so:

```
echo Model_Post::active_record()->name;
echo Model_User::active_record()->get_full_name();
```

There is almost no code involved in this, but I make sure to put it into every fuel project I work on.  It makes working on projects like this extremely fluid.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -80,6 +80,12 @@ class Model implements \ArrayAccess, \Iterator
 	protected static $_to_array_exclude = array();
 
 	/**
+	 * @var	array	Stores the all of the active records of the various classes
+	 */
+	
+	protected static $_active_record = array();
+
+	/**
 	 * @var  array  cached tables
 	 */
 	protected static $_table_names_cached = array();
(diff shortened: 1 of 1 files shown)


classes/model.php
```diff
@@ -214,6 +220,33 @@ public static function cached_object($obj, $class = null)
...
+	 * 
+	 * @access public
+	 * @static
+	 * @return self
+	 */
+	static function active_record(){
+		$c=get_called_class();
+		
+		// I am not sure whether this should return false, a blank class, or return false;
+		if (@!self::$_active_record[$c]) return $c::forge();
```
AddoSolutions: How this should return is a big question mark for me though.  Maybe a configuration option?

frankdejonge: This should have an isset check, not an `@` ... the @ is evil

WanWizard: using `self` is almost as evil. Use `static` instead.

WanWizard:
If you want a previously loaded object, it's probably best to just get it from cache. I can see the use-case for that, I'll put it on the list for the 2.0 ORM.

PR #276, opened 2013-06-29
Adding Hard Delete Functionality

craighooghiem:
I'm not sure if this is the best way to do it, but I think it's worth having even though the soft model is basically in stand-by mode until after the refactoring. We need a way to still delete records if we don't want them maintained. 

For the time being, this seems to work. I didn't have a chance to test all the cascades, etc but I see no reason why it wouldn't work exactly as expected. 

Function name is up for debate.

### DIFF
--- a/classes/model/soft.php
+++ b/classes/model/soft.php
@@ -263,9 +263,9 @@ class Model_Soft extends Model
 	 *
 	 * @return boolean
 	 */
-	public function delete_forever($cascade = null, $use_transaction = false)
+	public function purge($cascade = null, $use_transaction = false)
 	{
-		parent::delete();
+		parent::delete($cascade, $use_transaction);
 	}
 
 	protected function should_cascade_delete($rel)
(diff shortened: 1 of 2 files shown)


classes/model/soft.php
```diff
@@ -255,6 +255,19 @@ public function delete($cascade = null, $use_transaction = false)
...
+	 * Permanently deletes records using the parent Model delete function
+	 *
+	 * @param $cascade         boolean
+	 * @param $use_transaction boolean
+	 *
+	 * @return boolean
+	 */
+	public function delete_forever($cascade = null, $use_transaction = false)
+	{
+		parent::delete();
```
emlynwest: The two parameters, `$cascade` and `$use_transaction` should be passed to the parent call. I think calling the method `purge` would work better too, as that is what is used in Temporal iirc.

craighooghiem: Absolutely. My eyes clearly didn't see that.

Issue #279, opened 2013-07-02
to_array() on dump_tree(true) does not cascade


philipptempel:
Performing a `to_array()` on a result from `dump_tree(true)` does not cascade i.e., something like this

``` php
$roles = \Model\Role::forge()
    ->root()
    ->get_one()
    ->dump_tree(true);

\Debug::dump($roles->to_array());
```

results in an array for the root-node but the nodes below 'children' (and their children) are still all of type Orm\Model_Nestedset e.g.,

``` php
Variable #1:
  (Array, 8 elements) ↵
     children (Array, 3 elements) ↵
         4741165356817169 (Object #63): Orm\Model_Nestedset ↵

         8467357664215735 (Object #65): Orm\Model_Nestedset ↵

         4296539178188649 (Object #67): Orm\Model_Nestedset ↵

     resource_id (Integer): 2971968431872421
     name (String): "ROOT" (4 characters)
     slug (String): "" (0 characters)
     created_at (String): "1372068984" (10 characters)
     updated_at : null
     lft (String): "1" (1 characters)
     rgt (String): "8" (1 characters)
```

This also means, that `dump_tree(false)` does not yield the same result as `dump_tree(true)->to_array()`, as you'd expect, as long as I'm not mistaking the design of the model itself ;)

I'm trying to get my head around making a pull request, but I don't know where to insert the fix properly to not just fix this problem but also naut to cause any other problems ;)

WanWizard:
PR was merged.

PR #280, opened 2013-07-02
Update Orm\Model::to_object() to allow same parameters as to_array()

philipptempel:
I just ran across this case that did not support getting custom properties within a stdClass-object when running `to_object()` on an Orm\Model.

This fix introduces the same arguments `to_object($custom = false, $recurse = false)` as there are for `to_array()`. This allows for getting custom properties into the stdClass object representation of an Orm-object.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1915,9 +1915,9 @@ class Model implements \ArrayAccess, \Iterator
 	 *
 	 * @return  object
 	 */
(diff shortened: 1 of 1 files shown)


philipptempel:
This one should actually get a see #279 since they go hand in hand, however, this fix should be accepted with less code-fixing and code-breaking ;)

Issue #263, opened 2013-05-23
Can't delete model of Model_Soft has $_belongs_to property


takyam-git:
On 1.7/develop.

create 3 tables and models

``` php
//TABLES
\DBUtil::create_table('testchildren', array(
    'id' => array('constraint' => 11, 'type' => 'int', 'auto_increment' => true, 'unsigned' => true),
    'name' => array('constraint' => 255, 'type' => 'varchar'),
    'testparent_id' => array('constraint' => 11, 'type' => 'int', 'unsigned' => true), //NOT NULL
    'testpp_id' => array('constraint' => 11, 'type' => 'int', 'unsigned' => true), //NOT NULL
    'created_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
    'updated_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
    'deleted_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
), array('id'));
\DBUtil::create_table('testparents', array(
    'id' => array('constraint' => 11, 'type' => 'int', 'auto_increment' => true, 'unsigned' => true),
    'name' => array('constraint' => 255, 'type' => 'varchar'),
    'created_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
    'updated_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
    'deleted_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
), array('id'));
\DBUtil::create_table('testpp', array(
    'id' => array('constraint' => 11, 'type' => 'int', 'auto_increment' => true, 'unsigned' => true),
    'name' => array('constraint' => 255, 'type' => 'varchar'),
    'created_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
    'updated_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
    'deleted_at' => array('constraint' => 11, 'type' => 'int', 'null' => true),
), array('id'));
```

``` php
// MODELS
class Model_Testchild extends \Orm\Model_Soft
{
    protected static $_properties = array(
        'id',
        'name',
        'testparent_id',
        'testpp_id',
        'created_at',
        'updated_at',
        'deleted_at',
    );

    protected static $_observers = array(
        'Orm\Observer_CreatedAt' => array(
            'events' => array('before_insert'),
            'mysql_timestamp' => false,
        ),
        'Orm\Observer_UpdatedAt' => array(
[...]

PR #281, opened 2013-07-08
model_soft::delete() probrems

takyam-git:
ｆｉｘ 2 probrems.
### probrem 1 : Model_Soft::delete() with belongs to field that not null and has foreign key.

https://github.com/fuel/orm/issues/263 <- this issue.

bd7b62e

Model_Parent and Model_Child are Model_Soft.
##### save a parent with a chlid.

``` php
<?php
$parent = Model_Parent::forge(array('name' => uniqid()));
$parent->children[] = Model_Child::forge(array('name' => uniqid()));
$parent->save(true, true);
```
##### case 1: delete parent

``` php
<?php
$parent = Model_Parent::find('first');
$parent->delete(true, true); //success!!
/*
mysql> select * from parents;
+----+---------------+------------+------------+------------+
| id | name          | created_at | updated_at | deleted_at |
+----+---------------+------------+------------+------------+
|  1 | 51da8cb0b1852 | 1373277360 | 1373277411 | 1373277411 |
+----+---------------+------------+------------+------------+

mysql> select * from children;
+----+---------------+-----------+------------+------------+------------+
| id | name          | parent_id | created_at | updated_at | deleted_at |
+----+---------------+-----------+------------+------------+------------+
|  1 | 51da8cb0b1c66 |         1 | 1373277360 | 1373277411 | 1373277411 |
+----+---------------+-----------+------------+------------+------------+
*/
```
##### case 2: delete child

``` php
<?php
$child = Model_Child::find('first');
$child->delete();
/*
Fuel\Core\Database_Exception [ 1048 ]: Column 'parent_id' cannot be null [ UPDATE `children` SET `deleted_at` = 1373278091, `parent_id` = null, `updated_at` = 1373278091 WHERE `id` = '1' ]
*/
```
### problem 2: Model_Soft cascade delete Model

d0f6c69

case `Model_Parent` extends `\Orm\Model_Soft`, and `Model_Child` extends `\Orm\Model`.

``` php
<?php
$parent = Model_Parent::find('first');
$parent->delete(true, true);
/*
Orm\RelationNotSoft [ Error ]: Both sides of the relation must be subclasses of Model_Soft or Model_Temporal if cascade delete is true. Model_Child was found instead.
*/
```

---

This pull-request fix these 2 problems, but I don't know extent of the impact.

### DIFF
--- a/classes/belongsto.php
+++ b/classes/belongsto.php
@@ -203,19 +203,21 @@ class BelongsTo extends Relation
 		}
 	}
 
-	public function delete($model_from, $model_to, $parent_deleted, $cascade)
+	public function delete($model_from, $model_to, $parent_deleted, $cascade, $break_relation = true)
 	{
 		if ($parent_deleted)
 		{
 			return;
 		}
 
-		// break current relations
-		$model_from->unfreeze();
-		$rels = $model_from->_relate();
-		$rels[$this->name] = null;
-		$model_from->_relate($rels);
-		$model_from->freeze();
+		if($break_relation){
+			// break current relations
+			$model_from->unfreeze();
+			$rels = $model_from->_relate();
+			$rels[$this->name] = null;
+			$model_from->_relate($rels);
+			$model_from->freeze();
+		}
 
 		$cascade = is_null($cascade) ? $this->cascade_delete : (bool) $cascade;
 		if ($cascade and ! empty($model_to))
--- a/classes/model/soft.php
+++ b/classes/model/soft.php
@@ -193,7 +193,7 @@ class Model_Soft extends Model
 			$this->freeze();
 			foreach($this->relations() as $rel_name => $rel)
 			{
-				$rel->delete($this, $this->{$rel_name}, false, is_array($cascade) ? in_array($rel_name, $cascade) : $cascade);
+				$rel->delete($this, $this->{$rel_name}, false, is_array($cascade) ? in_array($rel_name, $cascade) : $cascade, false);
 			}
 			$this->unfreeze();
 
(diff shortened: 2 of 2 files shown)


steadweb:
:+1:

Thanks @takyam-git this patch worked for me, I was experiencing "RelationNotSoft" exception being thrown, where some of our models have lots of many_many relations, which may not extend Model_Soft / Model_Temporal.

WanWizard:
I don't agree with changes in the core Belongs_to class. And I don't think they are needed. Instead, the design of Model_Soft has to be looked at.

If a model has a belongs_to relation, the FK is always in the model itself. And since you are deleting that, there is no need to set anything to NULL.

takyam-git:
so, when create new model that belongs to(example's Model_Child), 
if allow set null to `parent_id`,  then we can be create not relation model.

``` php
<?php
$child = Model_Child::forge();
// $child->parent = $parent <= I forgot add this row !
$child->name = 'hoge';
$child->save();
```

I think, when allow null, it can be create bugs easily. The RDBMS can protect from bug of insert data, by `NOT NULL`.

takyam-git:
@WanWizard 
Soft deleted child can't be restore, because child record's paren_id is null.
Why set null to parent_id field on child's table ?

And legacy version ORM didn't set null to parent_id field.

``` sql
-- 1.5/develop:
UPDATE `children` SET `deleted_at` = '2013-07-24 05:51:50', `updated_at` = '2013-07-24 05:51:50' WHERE `id` = '38'
```

but now

``` sql
-- 1.7/develop:
UPDATE `children` SET `deleted_at` = '2013-07-24 05:52:14', `parent_id` = null, `updated_at` = '2013-07-24 05:52:14' WHERE `id` = '39'
```

These are same PHP code.

``` php
$model->delete(null, true);
```

I'm sorry my English If you feel a strange.

WanWizard:
I don't know why the parent_id is now included, but if that is the problem, that should be fixed. And not worked around...

PR #282, opened 2013-07-21
Fix bug in Orm\Model_Nestedset where primary-key was accessed as 'id'

philipptempel:
Line 1427 referenced primary-key named 'id' even though this must not be true for every model configuration i.e., did not support custom primary-keys.


WanWizard:
Hmm... Thought I'd fixed those? Thanks for spotting it.

PR #284, opened 2013-08-12
Add or_where case to _parse_where_array()

hackoh:
This change makes the following forms possible.

``` php
$items = \Model_Item::find('all', array(
    'where' => array(
        array(
            'or:0' => array('title', 'like', '%foo%'),
            'or:1' => array('description', 'like', '%foo%'),
        ),
        array(
            'or:0' => array('title', 'like', '%bar%'),
            'or:1' => array('description', 'like', '%bar%'),
        )
    )
));
```

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -558,7 +558,7 @@ class Query
 			if (is_array($v_w) and ! empty($v_w[0]) and is_string($v_w[0]))
 			{
 				! $v_w[0] instanceof \Database_Expression and strpos($v_w[0], '.') === false and $v_w[0] = $base.$v_w[0];
-				call_user_func_array(array($this, ($k_w === 'or' ? 'or_' : '').'where'), $v_w);
+				call_user_func_array(array($this, ($k_w === 'or' || strpos($k_w, 'or:') === 0 ? 'or_' : '').'where'), $v_w);
 			}
 			elseif (is_int($k_w) or $k_w == 'or')
 			{


WanWizard:
@stevewest I wonder if we should put more effort in this. 

I'm seriously contemplating dropping this array notation for 2.0, it is utterly complex, and very inflexible as you can't deal with order and grouping without resorting to hacks like this...

This is a lot easier, and everyone can understand it:

```
$items = \Model_Item::query()
    ->where_open()
        ->where('title', 'like', '%foo%')
        ->or_where('description', 'like', '%foo%')
    ->where_close()
    ->or_where_open()
    ->where_open()
        ->where('title', 'like', '%bar%')
        ->or_where('description', 'like', '%bar%')
    ->where_close()
    ->get();
```

At least, that is what I think this syntax is supposed to do. If not, that just highlights another problem with this array, it's complexity.

hackoh:
I think, It is not very desirable to limit an order of a method call. 
So, I do not like building the query by "method chain" pattern..

emlynwest:
@WanWizard I think that limiting the interface to method chaining would be for the best. I have spent lots of time trying to work out what the exact structure of the undocumented array should be ;)

@hackoh By using method chaining it allows for easier hints for the dev. Using the methods mean that things like auto-completion, it removes the need to guess at the array structure and the random errors that result from getting it wrong.

hackoh:
I do not think that it is a correct answer to generate a query by the array, too. 
But, I illustrate a little example below. 

``` php

Model_Article::find('all', array(
    'filter' => Filter::and_link(array(
        Filter::equal('is_enable', true),
        Filter::and_link(array(
            Filter::or_link(array(
                Filter::both_like('keyword', 'foo'),
                Filter::both_like('content', 'foo'),
            )),
            Filter::or_link(array(
                Filter::both_like('keyword', 'bar'),
                Filter::both_like('content', 'bar'),
            )),
        ))
    ))
));

```

```
Filter_Link_And -> Filter_Link -> Filter
```

``` php
abstract class Filter_Link extends Filter implements ArrayAccess
{
...
}
```

This syntax can be very sensed. 
It is possible to make shortcut, if you do not like the redundant syntax of 'Filter'.
For example,

``` php
Model_Article::find('all', array(
    'filter' => __and(array(
        __equal('is_enable', true),
        __and(array(
            __or(array(
                __both_like('keyword', 'foo'),
                __both_like('content', 'foo'),
            )),
            __or(array(
                __both_like('keyword', 'bar'),
                __both_like('content', 'bar'),
            )),
        ))
    ))
));

```

First of all, naming of the argument 'where' is poisoned to the curse of SQL.

hackoh:
'Filter' should be ignorant to SQL. 
And, a certain class belonging to the context of Orm\Model should have the capability to parse Filter.

emlynwest:
So if you are doing that what's wrong with method chaining?

hackoh:
Generation of a query with the method chain which restricted a call order is visible to the pattern which utilized the object,But, the actual condition is Procedural programming. 
The design of application is also dragged by it..

In 1.x system, I do not want the above-mentioned demand to be drunk. 
However, in an inheritor, I want you not to fall into an anti pattern called a method chain. 
Therefore, I shout this.

WanWizard:
It hasn't been decided yet what the interface of the 2.0 ORM will be. But as far as I'm concerned, the current array notation goes. It's limiting, confusing, and undocumented.

1.x is slowly approaching it's end, it's pointless making such changes in the correct code. So we won't.

hackoh:
It is satisfactory to throw away this PR. 
However, I hope that you refer to the notation written by me by https://github.com/fuel/orm/pull/284#issuecomment-22502160 for 2.0    :)

WanWizard:
I don't really like both suggestions, but we'll see what we end up with.

PR #285, opened 2013-08-12
fix Model_Soft::restore() doesn't work well, because belongs_to field is null on delete.

takyam-git:
other way of https://github.com/fuel/orm/pull/281

### DIFF
--- a/classes/belongsto.php
+++ b/classes/belongsto.php
@@ -210,12 +210,16 @@ class BelongsTo extends Relation
 			return;
 		}
 
-		// break current relations
-		$model_from->unfreeze();
-		$rels = $model_from->_relate();
-		$rels[$this->name] = null;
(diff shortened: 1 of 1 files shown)


emlynwest:
Looks like a workable solution but I don't like the idea of having the model specific stuff in the relation classes.

@WanWizard Would it be worth using this as a temp fix until the v2 orm is done?

takyam-git:
So but fix with flagment on [my pull request (#281)](https://github.com/fuel/orm/pull/281) is not merged.

https://github.com/takyam-git/orm/commit/bd7b62eef6ea5cf416ec4c17cac18318453c8745
https://github.com/takyam-git/orm/commit/8980caf4a7c1242afe0047b2bdf91ff75b716c72

Do you have other solutions?

PR #287, opened 2013-08-20
Fix bug in Orm\Model_Nestedset::dump_tree()

feldsam:
Whem node had 2 and more levels of childs, next node on level 0 was child of one level up node in returned array.


WanWizard:
Very well spotted. Thanks for the fix!

PR #289, opened 2013-08-25
Removed int casting for min/max to allow for dates

jraede:
The max and min methods casted the result to an integer before returning - I don't really see a reason for this as the user should be aware of the data type on the database, and casting to an int prevents retrieval of min/max for dates and datetimes.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -1341,7 +1341,7 @@ class Query
 			return false;
 		}
 
-		return (int) $max;
+		return $max;
 	}
 
 	/**
@@ -1376,7 +1376,7 @@ class Query
 			return false;
 		}
 
-		return (int) $min;
+		return $min;
 	}
 
 	/**


WanWizard:
@FrenkyNet @stevewest opinions? It might create a BC issue...

emlynwest:
Personally I think it's something we should leave be for now and can address in/will be removed with the v2 orm. Does the typing observer not already do this?

jraede:
From what I'm seeing the typing observer doesn't get called here because it's not assigning any properties on a specific model and thus not triggering a model event

frankdejonge:
@WanWizard I believe this was added not too long ago and the reason to remove it is very valid, dates should not be cast to integers. So it's got my +1

emlynwest:
Looking at this again after getting some sleep I agree with @FrenkyNet :+1:

WanWizard:
I checked the code, it was added 6 months ago. See https://github.com/fuel/orm/pull/231 (you're contradicting yourself, @FrenkyNet :-))

frankdejonge:
@WanWizard I'm not, we (all) didn't think of the max/min being used on date-time stuff, so this brings new conditions into play. Hence a different possible outcome, which in this case is called for.

WanWizard:
lol, just kiddin' ;-)

I agree that this changes it, the previous PR clearly broke BC, much more than this will do now...

PR #290, opened 2013-08-25
Pagination refactor

jraede:
ORM query pagination has always had issues when adding relations to the mix. This fixes that by avoiding the use of LIMIT and OFFSET, and rather just adding one extra query to retrieve the primary keys of all matching main (first-level) models, and then adding a where condition for {PRIMARY_KEY} IN ($values), where $values is the paginated set of primary key results. 

So basically, no more need for subqueries, it functions the same as \Orm\Query::count() in terms of query generation, and just applies pagination based on the desired position in the full primary key result set.

Example:

```
$query = \Model\Paycheck::query()->related('employee')->related('company')->where('company.city', 'Los Angeles');

// This is optional - if you don't provide an instance of \Fuel\Core\Pagination then the add_pagination method
// will create one on the variable passed by reference to the method (third argument)
$pagination = \Pagination::forge('my-pagination', array(
    'pagination_url'=>'/paycheck/index', 
    'uri_segment'=>'page'
));

$results = $query->order_by('employee.name')->apply_pagination(\Input::get('page', 1), 20, $pagination)->get();
```

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -1357,7 +1357,10 @@ class Query
 		// Should be already ordered.
 		$values = array_unique($this->values());
 		$total = count($values);
-
+		if(!$total) {
+			return $this;
+		}
+		
 		$config = array(
 			'total_items'    => $total,
 			'current_page'=>$page,
@@ -1417,7 +1420,7 @@ class Query
 			return false;
 		}
 
-		return (int) $max;
+		return $max;
 	}
 
 	/**
@@ -1452,7 +1455,7 @@ class Query
 			return false;
 		}
 
-		return (int) $min;
+		return $min;
 	}
 
 	/**
--- a/classes/query.php
+++ b/classes/query.php
@@ -1407,7 +1407,7 @@ class Query
 			return false;
 		}
 
-		return $max;
+		return (int) $max;
 	}
 
 	/**
@@ -1442,7 +1442,7 @@ class Query
 			return false;
 		}
 
-		return $min;
+		return (int) $min;
 	}
 
 	/**
--- a/classes/query.php
+++ b/classes/query.php
@@ -1323,8 +1323,9 @@ class Query
 	 * and it's much more complicated to add offset and limit. For example, you get the total, get the IDs, apply offset + limit
 	 * with PHP, and then add where('id', 'IN', $ids) to the final query to do "bogus" pagination
 	 * 
-	 * @param  [type] $column [description]
-	 * @return [type]         [description]
+	 * @param  string $column Column to get the value(s) of. Defaults to primary key
+	 * @return array array of values
+	 * @author  Jason Raede <jason.raede@gmail.com>
 	 */
 	public function values($column = null) {
 		$select = $column ?: \Arr::get(call_user_func($this->model.'::primary_key'), 0);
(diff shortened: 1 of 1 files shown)


jraede:
Rereading my initial comment, it doesn't even create one extra query, it just replaces the typical use of \Orm\Query::count() to get the total with a different query to get all of the primary keys, so the added overhead is virtually nil as far as I can tell.

WanWizard:
It is not a good idea to mix functionalities, and pagination has no business in an ORM.

The ORM has been designed to make sure resultsets are always complete, which means you can't limit on a specific number, since the definition of "complete" depends on the number of related records.

As the docs state, if you run a query with LIMIT/OFFSET that includes related items, you have to use the rows_limit() and rows_offset() methods, that were introduced to bypass this design, especially for pagination purposes.

jraede:
Right, but if you have a where condition on a relation, that where condition does not go into the subquery generated by \Orm\Query when offset/limit is applied, so you get the wrong result set. If pagination is not meant to be applied to the ORM package at all, then I see your point, but seeing as you already have functionality built in to accommodate pagination that isn't fully working when you add in relations, I think this is a viable fix.

Example:

```
$query = \Model\Paycheck::query()->related('employee')->related('employee.notes')
    ->related('employee.company')->where('employee.company.city', 'Los Angeles')
    ->limit(20)->offset(0)
```

Will yield something like this:

```
SELECT t0.*, t1.*, t2.*, t3.* FROM (SELECT * FROM paychecks LIMIT 20 OFFSET 0) 
AS t0 LEFT JOIN employees AS t1 ON (t0_c1 = t1_c0) LEFT JOIN notes AS t2 ON (t2_c1 = t1_c0) 
LEFT JOIN companies AS t3 ON (t3_c0 = t1_c1) WHERE t3_c2 = 'Los Angeles'
```

However, that FIRST limits the paychecks to 20, and THEN applies the where company.city = 'Los Angeles', so you would get a blank result set if you're querying the first page but there are no related companies in Los Angeles for the first 20 paychecks.

Alternatively, as you suggest, using rows_offset and rows_limit would yield something like this:

```
SELECT t0.*, t1.*, t2.*, t3.* FROM paychecks AS t0 LEFT JOIN employees AS t1 ON (t0_c1 = t1_c0) 
LEFT JOIN notes AS t2 ON (t2_c1 = t1_c0) LEFT JOIN companies AS t3 ON (t3_c0 = t1_c1) 
WHERE t3_c2 = 'Los Angeles' LIMIT 20 OFFSET 0
```

But notes, a has_many relation, is included in the query. This will not just limit the number of paychecks but rather it will limit the total records, which could leave you with 10 paychecks and 10 notes, or even 1 paycheck and 19 notes, instead of the desired 20 paychecks.

The way around this is in the pull request - don't use LIMIT and OFFSET clauses, but rather use PHP to explicitly limit the result set to 20 paychecks no matter what.

WanWizard:
Ah, wait a second.

Your understanding / requirements of pagination are completely different.  For pagination, rows_limit()/rows_offset(), will do the trick just nicely, they return exactly the about of requests requested, like the original intention of the LIMIT and OFFSET statements.

This is however not what you want. You DO want the limit/offset on the subquery (to get the 20 paychecks you want). Your issue is that the WHERE clause is positioned incorrectly.

Some of the issues related to the subquery and additional where's are fixed in the 1.7/develop branch, but some are still open, and are difficult to fix due to the way the query is generated. They might not get fixed before 2.0, which will get a complete rewrite of the ORM.

jraede:
Gotcha. Well like you said it's difficult to fix because of all of the different possible combinations of relations and wheres and subqueries and all of that - why not just avoid using pagination via LIMIT and OFFSET and just apply it before hand? From my testing, the change I put in accounts for all of those situations without any added overhead.

WanWizard:
The entire subquery business needs to be looked at and improved for 2.0

Issue #288, opened 2013-08-21
Model_Soft::count()/max()/min() result include deleted rows.


takyam-git:
When a table `foos` has 2 alive rows and 2 deleted rows, `Model_Foo::count()` expect result is `2`, but I got `4`, this method include deleted rows.

https://github.com/fuel/orm/commit/6222bb9795d38cdd98b7849d775757230c041555

In this commit, process that set query `deleted_at is NULL` move from `set_soft_filter()` to `get()`, but `count()`, `max()` and `min()` aren't use `get()` method.

takyam-git:
thanks!

PR #291, opened 2013-08-26
Orm/Query_Soft::count()/max()/min() arguments missing

takyam-git:
I got notice error.

``` text
Runtime Notice!
Fuel\Core\PhpErrorException [ Runtime Notice ]:
Declaration of Orm\Query_Soft::count() should be compatible with Orm\Query::count($column = NULL, $distinct = true)
```

related issue #288


WanWizard:
Well spotted, thanks!

PR #292, opened 2013-08-29
Remove one too many closing bracket on L1250

philipptempel:
There was one too many closing bracket in orm/model/nestedset.php on line 1250 that broke code.


WanWizard:
Now that was a stupid mistake... :-(

PR #299, opened 2013-09-18
Bugfix : hasmany relations correctly hydrated when using from_cache=false

felixgilles:
Without this commit, each item of results only contains the first related item.

My proposition: in hydrate, try to retrieve object in the current results before trying to retrieve it from cache.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -1068,9 +1068,21 @@ class Query
 			$obj[$pk] = $row[$pk_c];
 		}
 
-		// Check for cached object
+		// Check for already builed object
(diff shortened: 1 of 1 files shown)


felixgilles:
Hi @WanWizard,

Is it the suppression of the 1.7/develop branch, when you create 1.7/master and  1.8/develop, that have closed  automatically this PR and #305 ?

I reopens them on 1.8/develop ?

WanWizard:
I was not aware that removing a branch would automatically close PR's, I never noticed that before.

Issue #300, opened 2013-09-18
Indirect modification of overloaded property Model_Nestedset::$children has no effect


sagikazarmark:
This error is thrown, when I try to call dump_tree function with as_object set to true.

In this model EAV is set. The problem is that the eav getter does not return by reference, but by value, so the property does not exists. (UPDATE: the getter DOES return by reference, but it is not a reference of the object in case of EAV.)

sagikazarmark:
Possible solution: return _get_eav by reference. I don't know if that has any effect on other things, but solves this one.

sagikazarmark:
No, it is only a partial solution as this way an error will be thrown, that the property does not exists.

sagikazarmark:
I am not sure, I understand what I did, but I am sure that this problem is related to the fact, that _get_eav does not return by reference.

My PHP version is 5.4.9.

Actually, EAV is not working with Nestedsets at all, but also throws this error. Is this the expected functionality? (I mean, is EAV with Nestedsets implemented? should it work?)

sagikazarmark:
Possible solution: $children and the $path values from dump_tree could be stored in an "exclusion" array, so that properties are always returned by reference.

sagikazarmark:
@WanWizard can you help me understand this issue? I am a bit confused now. :smile:

WanWizard:
Is this still an issue? 

As far as I can see, there is nothing in dump_tree() that could interfere with EAV, nothing accesses properties in that method, only Model object references are assigned to a tree structure.

sagikazarmark:
I will check it out.

WanWizard:
Any results from this checking?

PR #302, opened 2013-09-21
Model_Nestedset dump_tree() parameters ($children and $path) are set and get by reference

sagikazarmark:


sagikazarmark:
This worked, but I found an other bug, but this is supposed to solve #300.

sagikazarmark:
Ok, this should work now. Once again: this way you can dump a tree as objects, because $children and $path are set by reference of the object, and EAV don't mess it up.

sagikazarmark:
Did you check this out?

sagikazarmark:
@stevewest can you check this out?

PR #310, opened 2013-10-16
Implement ability to do get-time sanitization

phil-lavin:


phil-lavin:
Releated to https://github.com/fuel/core/pull/1548

WanWizard:
Sanitize data on a get() is imho a very bad idea. How are you going to get unaltered data out?

phil-lavin:
@WanWizard it doesn't differ a lot to how it works now. At present, the whole model is sanitized and its original data is changed before the view is rendered. In this usecase, instead of sanitizing the whole model, the model is marked as ready to be sanitized. The result of get is the same either way.

Also, annoyingly, at present if you save a model which has been passed to a rendered view already, you save the sanitized data to the DB.

WanWizard:
I think it should work as notification, with an on/off switch. The security class can switch it on when it detects the interface implemented (which is a simple method call), the class itself can then internally decide what to do different when the flag is on. 

I'll work something out.

WanWizard:
Noticed this PR is still open. This has been implemented in the Security class and for ORM models some time ago.

PR #315, opened 2013-10-21
Adds the ability to include relations when using the UpdatedAt observer

emlynwest:
Putting this up here to get thoughts on the feature/implementation. I thought this would be a useful feature as often an ORM entity consists of multiple models.

My example is that a blog post has many "i18ns", translatable data stored in a separate table (so another model), when any of the i18ns are updated I want to be able to reflect that change on the parent blog post. This observer will allow this to work.

``` php
$blog_post->i18ns[$i18n_id]->title = 'My new title';
$blog_post->save();
// $blog_post->updated_at is now "now()"
```

The added setting is used like this:

``` php
    'Orm\\Observer_UpdatedAt' => array(
        'events' => array('before_save'),
        'mysql_timestamp' => true,
        'property' => 'updated',
        'relations' => array(
                'i18ns',
        ),
    ),
);
```

Docs are on the way.

### DIFF
--- a/classes/observer/updatedat.php
+++ b/classes/observer/updatedat.php
@@ -116,7 +116,16 @@ class Observer_UpdatedAt extends Observer
 			return false;
 		}
 
-		// Now we know the relation exists loop through all the models and see if anything has changed
+		$relation_object = $obj->relations($relation);
+
+		// Check if whe have a singular relation
+		if ($relation_object->is_singular())
+		{
+			// If so check that one model
+			return $obj->{$relation}->is_changed();
+		}
+
+		// Else we have an array of related objects so start checking them all
 		foreach ($obj->{$relation} as $related_model)
 		{
 			if ($related_model->is_changed())
--- a/classes/observer/updatedat.php
+++ b/classes/observer/updatedat.php
@@ -83,7 +83,7 @@ class Observer_UpdatedAt extends Observer
 			if ($this->relation_changed($obj, $relation))
 			{
 				$relation_changed = true;
-				continue;
+				break;
 			}
 		}
 
@@ -108,6 +108,12 @@ class Observer_UpdatedAt extends Observer
 		if ($obj->relations($relation) === false)
 		{
 			throw new \InvalidArgumentException('Unknown relation '.$relation);
+		}
+
+		// If the relation is not loaded then ignore it.
+		if ( ! $obj->is_fetched($relation))
+		{
+			return false;
 		}
 
 		// Now we know the relation exists loop through all the models and see if anything has changed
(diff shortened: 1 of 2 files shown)


WanWizard:
docs ? ;)

emlynwest:
As usual something exploded and needed to be fixed. I'll try and get the PR for the docs up today.

classes/observer/updatedat.php
```diff
@@ -70,9 +76,49 @@ public function before_save(Model $obj)
...
 	{
-		if ($obj->is_changed())
+		// If there are any relations loop through and check if any of them have been changed
+		$relation_changed = false;
+		foreach ( $this->_relations as $relation)
+		{
+			if ($this->relation_changed($obj, $relation))
+			{
+				$relation_changed = true;
+				continue;
```
andrewwheal: Would this not want to be break to avoid unnecessarily checking further relations?

classes/observer/updatedat.php
```diff
@@ -70,9 +76,49 @@ public function before_save(Model $obj)
...
+	protected function relation_changed(Model $obj, $relation)
+	{
+		// Check that the relation exists
+		if ($obj->relations($relation) === false)
+		{
+			throw new \InvalidArgumentException('Unknown relation '.$relation);
+		}
+
+		// Now we know the relation exists loop through all the models and see if anything has changed
+		foreach ($obj->{$relation} as $related_model)
```
andrewwheal, line 129: This method of accessing a relation on a model would lazy load it from the database if it has not already been accessed/loaded.

If the relation has not been accessed then it should be safe to assume that it has not changed. If the relation could be accessed/checked via the Orm\Model internals then it could avoid unnecessary DB calls.

andrewwheal, line 129: Could `Orm\Model::is_changed($relation)` be used to do this checking?

emlynwest, line 129: `Orm\Model::is_changed($relation)` only checks if models have been added or removed from the relation.

emlynwest:
fuel/docs#599 for docs.

classes/observer/updatedat.php
```diff
@@ -70,9 +76,55 @@ public function before_save(Model $obj)
...
+		}
+
+		// If the relation is not loaded then ignore it.
+		if ( ! $obj->is_fetched($relation))
+		{
+			return false;
+		}
+
+		// Now we know the relation exists loop through all the models and see if anything has changed
+		foreach ($obj->{$relation} as $related_model)
```
andrewwheal, line 129: There needs to be a check for singular relations as they would break this loop

emlynwest:
Just need to fix an issue of not being able to handle has_one or belong_to

PR #316, opened 2013-10-22
Allows soft models w/slug observer to be deleted+recreated w/same source property

chaddoncooper:
Issue can be re-created with a fresh install of 1.8 develop:

Create a Soft Model:

``` php
class Model_Test extends Orm\Model_Soft
{
    protected static $_properties = array(
        'id',
        'title',
        'slug',
        'deleted_at',
    );


        'Orm\\Observer_Slug' => array(
            'events' => array('before_insert', 'before_update'),
            'source' => 'title',
            'property' => 'slug',
        ),
    );
}
```

Run a controller method with the following:

``` php
        DBUtil::truncate_table('tests');

        $model = Model_Test::forge(array(
            'title' => 'a',
            )
        );

        $model->save();

        $model->delete();

        $model = Model_Test::forge(array(
            'title' => 'a',
            )
        );

        $model->save();
```

Exception will be thrown on duplicated row:
    Fuel\Core\Database_Exception [ 1062 ]:
    Duplicate entry 'a' for key 'slug' [ INSERT INTO `tests` (`title`, `slug`, `deleted_at`) VALUES ('a', 'a', null) ]

I have mitigated the issue by adding a boolean property to the slug observer class which disables the filter temporarily for soft models, while it builds the query to search for duplicate properties (title by default).

### DIFF
--- a/classes/observer/slug.php
+++ b/classes/observer/slug.php
@@ -30,6 +30,11 @@ class Observer_Slug extends Observer
 	public static $property = 'slug';
 
 	/**
+	 * @var  string  Default disable_soft property, allows Model_Soft objects to be deleted and created with the same title
+	 */
+	public static $disable_soft = false;	
+	
+	/**
 	 * @var  mixed  Source property or array of properties, which is/are used to create the slug
 	 */
 	protected $_source;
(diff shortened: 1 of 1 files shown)


WanWizard:
Isn't the issue here that you have a unique index on "slug", which with soft-delete is not always the case (as you have noticed)?

Records marked deleted should be treated as such, so they should not be taken into account in the observer. If you have a column that should be unique (amongst all non-deleted records), you should take care of this when un-deleting a deleted record.

@stevewest, your opinion?

emlynwest:
This was originally my fix ( @chaddoncooper is a minion) but since doing this I feel you are right, the unique index should be removed.

PR #319, opened 2013-10-31
Orm\Query::get_query() fixed ignore model connection setting

feldsam:
This was issue, if defined connection has different table preffix that default connection.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -1238,7 +1238,7 @@ class Query
 		// Build the query further
 		$tmp     = $this->build_query($query, $columns);
 
-		return $tmp['query'];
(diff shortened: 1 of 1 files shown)


WanWizard:
@feldsam what was the use-case for this? You changed the returned result from a DB object to a string.

feldsam:
Hi, sorry for breaking subqueries. Problem is if you have default db connection with table preffix and second connection with different prefix, them get_query() method generete query with table prefixes from default connection on model which use second connection .

WanWizard:
@feldsam This might fix it for you.

feldsam:
thank you

PR #320, opened 2013-11-08
Update model.php

57u:
commented out duplicate code. i think it was causing "cannot update primary key" errors when deleting models

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1455,6 +1455,8 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 				return false;
 			}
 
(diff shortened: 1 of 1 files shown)


emlynwest:
If you look closer I think you will find there's a very minor difference.

57u:
I see. oops. 

Im still confused as to why when I’m deleting an object, orm is trying to update the primary key on a related object. the object in question _has_one_ (a content has a type), on content->delete() i get errors saying cannot update primary key on the type model….

emlynwest:
The ORM will update the FK of a related table to break that relation before deleting. If your FK happens to also be the PK for your table this will fail as you can not modify a PK once it's been set.

PR #322, opened 2013-11-30
Observer Slug added separator parameter

sagikazarmark:


sagikazarmark:
Documentation also sent

Issue #303, opened 2013-09-21
Dump EAV to array


sagikazarmark:
I am thinking about the implementation of dumping EAV attributes in to_array().

What do you think?

sagikazarmark:
@WanWizard can you check this? Actually I already implemented it, but I don't want to PR unless I know that this is not forbidden or something :smile:

sagikazarmark:
@stevewest can you check this out?

emlynwest:
EAV is not my domain, you will need @WanWizard's input for this. IMO the best thing would be to code it up and create a pull request then we can take a look at what you propose. Personally I don't use EAV much and have never had to use `to_array()` with it but I could see how it might be useful.

WanWizard:
EAV data is already available in to_array() output, as underneath it's a plain one-to-many relation?

PR #329, opened 2013-12-09
EAV properties added to to_array

sagikazarmark:
All properties of all related EAV relations are included in the array. EAV properties of relations are also included on the level of the relation.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1951,6 +1951,7 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 	 *
 	 * @param bool $custom
 	 * @param bool $recurse
+	 * @param bool $eav
 	 *
 	 * @internal param \Orm\whether $bool or not to include the custom data array
 	 *
@@ -2012,7 +2013,7 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 					static::$to_array_references[] = get_class(reset($rel));
 					foreach ($rel as $id => $r)
 					{
-						$array[$name][$id] = $r->to_array($custom, true);
+						$array[$name][$id] = $r->to_array($custom, true, $eav);
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Started in #303

sagikazarmark:
There should be a publicly visible way to get the eav relations. Use case: you want to dynamically load EAV relations of several models in a query and you don't know the name of eav relations.

WanWizard:
This can produce inconsistent output. 

When you have multiple EAV containers defined, __get() returns the first hit according to the order of the relations defined. In this code, the last hit wins.

And you still have the data in your array twice, as all EAV will also be in the array as related data. 

I'm not convinced of the use-case, you're the developer, how can you not know the name of the EAV relation?

sagikazarmark:
The first problem's solution is simple: only set the value if it is not in the array. Before getting the values from the relations I process the EAV containers in order they are defined. So the same field don't have to be overwritten.

I think NOT having the EAV fields in the array IS an inconsistent output. If I use the model as an object, either getting a property or ArrayAccess way I get the EAV property. After to_array, I don't?

I have the data twice, and I think that having EAV container relations in to_array makes almost no sense, but removing it can break things.

But if you don't know and don't want to know whether you are getting keys from an array or an ArrayAcces, and you cannot get the same data (the same way), that is the definition of inconsistent output. (Eg. Twig, you don't have to know whether it is an array/object, you can get properties the same way.)

Last but not least: think about a controller which gets data from ANY model you give it. Eg. a blog module. You have a module for every post type. You can't be sure of the EAV name, unless YOU wrote all the models. But this last thing is actually useless or at least specific to my blog module.

sagikazarmark:
If you don't like having the data twice, the I can make it a reference to the original array value for you. :smile:

sagikazarmark:
I can imagine a third property in to array which tells the function whether you want to eav properties like eav properties or not.

WanWizard:
I was about to suggest that:
- introduce a third parameter to indicate you want to include EAV or not (like $custom)
- make sure EAV relations are processed in reverse order, so the data will be the same as from get()
- filter the EAV relations from the included relations, no need to add the same data twice

sagikazarmark:
Won't the last point break BC?

sagikazarmark:
Any update on this?

PR #331, opened 2013-12-12
fixed many-to-many relation not using write connection

SlKelevro:
This already cost us some problems with replication =(

There are also `Orm\Model::connection()` calls in the `join()` method of each relation, but I'm not sure if they should be fixed too (seems like they're only used with SELECT queries).


WanWizard:
Good find! 

I did a check on all files in the package, but these three seem to be the only ones that would require the "write" connection.

PR #332, opened 2014-01-10
FEATURE: select custom sql columns

keigoi:
A new `Query` method `custom()` , which adds extra columns to select. Resulting model will have propertes for those columns.

Example:

``` php
// assume Model_Person has properties 'firstname' and 'sirname' but not 'fullname'.
$query = Model_Person::query();
$query->custom(array('fullname' => DB::expr('CONCAT(firstname,\' \',sirname)')));
$people = $query->get();
foreach($people as $id => $person)
{
    echo $person->fullname;
}
```

Adding columns in such way would be useful, for example:
- relevance score for full-text search
- row numbers for a table
- and other expressions better to calculate on database server

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -796,9 +796,9 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 			$new = false;
 		}
 
+		$properties = $this->properties();
 		if ($new)
 		{
-			$properties = $this->properties();
 			foreach ($properties as $prop => $settings)
 			{
 				if (array_key_exists($prop, $data))
@@ -816,7 +816,14 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 		else
 		{
 			$this->_update_original($data);
-			$this->_data = array_merge($this->_data, $data);
+			$obj = array();
+			foreach ($properties as $prop => $settings)
+			{
+				$obj[$prop] = $data[$prop];
+				unset($data[$prop]);
+			}
+			$this->_data = array_merge($this->_data, $obj);
+			$this->_custom_data = $data;
 
 			if ($view and array_key_exists($view, $this->views()))
 			{
--- a/classes/query.php
+++ b/classes/query.php
@@ -74,6 +74,11 @@ class Query
 	protected $select = array();
 
 	/**
+	 * @var  array  additional expressions to select, will be stored into $model->_custom_data
+	 */
+	protected $select_custom = array();
+
+	/**
 	 * @var  int  max number of returned base model instances
 	 */
 	protected $limit;
@@ -270,7 +275,13 @@ class Query
 			$out = array();
 			foreach($this->select as $k => $v)
 			{
-				$out[] = array($v, $k);
+				$out[] = array($v, $k, (strpos($v, '.') === false ? $v : substr($v, strpos($v, '.') + 1)));
+			}
+
+			$i = count($out);
+			foreach($this->select_custom as $k => $v)
+			{
+				$out[] = array($v, $this->alias.'_c'.$i++, $k);
 			}
 
 			// set select back to before the PKs were added
@@ -297,6 +308,24 @@ class Query
 			}
 		}
 
+		return $this;
+	}
+
+	/**
+	 * Add custom columns to select, which are included, each as its own param.
+	 * For instance, $query->custom(array('fullname' => DB::expr('CONCAT(firstname,\' \',sirname)'))) will
+	 * add an extra property 'fullname' into the resulting model.
+	 *
+	 * @param array  $columns    Extra columns to select
+	 *
+	 * @return  void|array
+	 */
+	public function custom($columns)
+	{
+		foreach ($columns as $colname => $expr)
+		{
+		   $this->select_custom[$colname] = $expr;
+		}
 		return $this;
 	}
 
@@ -1079,7 +1108,7 @@ class Query
 			$obj = array();
 			foreach ($select as $s)
 			{
-				$f = substr($s[0], strpos($s[0], '.') + 1);
+				$f = $s[2];
 				$obj[$f] = $row[$s[1]];
 				if (in_array($f, $primary_key))
 				{
@@ -1094,7 +1123,7 @@ class Query
 			// add fields not present in the already cached version
 			foreach ($select as $s)
 			{
-				$f = substr($s[0], strpos($s[0], '.') + 1);
+				$f = $s[2];
 				if ( ! isset($obj->{$f}))
 				{
 					$obj->{$f} = $row[$s[1]];


WanWizard:
@stevewest what do you think?

I have my doubts as to the usefulness of this feature. An ORM is not meant to be a query builder, the queries given in the use-case should be created using a standard DB query.

emlynwest:
@WanWizard I agree, I do not think this is something that should be included in the current ORM. Any changes to the `diff()` function are going to break BC as well as `Model_Temporal` as that relies on the `diff()` result to work out when to insert revisions.

keigoi:
Okay, I noticed my commit on `model.php` would break some ORM functions. Thank you guys for pointing it out. 
I would not hesitate to withdraw my request, then fix it and request again later.

However, I insist this feature is indeed useful in some aspects:
- This kind of 'escape hatch' will facilitate much usability of FuelPHP. 
- DB class provides much low-level feature. There's no relate() nor count().
- Last but not least, there's another urgent real-world need for this feature. Here's another example.

``` php
// Assume we are building a blog site. We have articles and their comments on separate tables.
// Here we are to implement fulltext search...

// MySQL specific feature: MATCH(columns) AGAINST ('search words') will return relevance score of that row.
$match_expr = DB::expr('MATCH (title,body) AGAINST ('.DB::quote($words).' IN BOOLEAN MODE)');

$query = Model_article::query()
    ->custom(array('search_score'=>$match_expr)) // add score row
    ->where(array($match_expr, null, null))
    ->order_by($match_expr, 'desc');

// here we use ORM's count(), limit() and offset() for standard pagination, 
// which can be much more tricky if we use 
// DB::select() and as_array() within reasonable computational cost.
$pagination = Pagination::forge('mydefault', array('total_items' => $query->count(),'name'=>'mydefault'));
$this->template->set_global('pagination', $pagination->render(), false);
$query->limit($pagination->per_page)->offset($pagination->offset);

// and more, thanks to ORM, here we can associate comments without any burden!
$articles = $query->related('comment')->get();

...
```

WanWizard:
The change in model.php actually addresses a bug, so that has to go in, but then so that it doesn't break anything. I'll do that in a minute.

Once that is done, there is no need for the rest anymore, since you can do

```
$result = DB::select()
    ->where(DB::expr('MATCH (title,body) AGAINST ('.DB::quote($words).' IN BOOLEAN MODE)')
    ->as_object('Model_Article')
    ->execute();
```

which you now can't because of the fact that the constructor doesn't support custom data.

WanWizard:
I'd say the constructor should be this:

``` php
    /**
     * Constructor
     *
     * @param  array
     * @param  bool
     */
    public function __construct(array $data = array(), $new = true, $view = null, $cache = true)
    {
        // This is to deal with PHP's native hydration that happens before constructor is called
        // for some weird reason, for example using the DB's as_object() function
        if( ! empty($this->_data))
        {
            // merge the injected data with the passed data
            $data = array_merge($this->_data, $data);
            $this->_data = array();

            // and mark it as existing data
            $new = false;
        }

        // move the passed data to the correct container
        $properties = $this->properties();
        foreach ($properties as $prop => $settings)
        {
            // do we have data for this this model property?
            if (array_key_exists($prop, $data))
            {
                // store it in the data container
                $this->_data[$prop] = $data[$prop];
                unset($data[$prop]);
            }

            // property not present, do we have a default value?
            elseif ($new and array_key_exists('default', $settings))
            {
                $this->_data[$prop] = $settings['default'];
            }
        }

        // store the remainder in the custom data store
        $this->_custom_data = $data;

        // store the view, if one was passed
        if ($view and array_key_exists($view, $this->views()))
        {
            $this->_view = $view;
        }

        if ($new === false)
        {
            // update the original datastore and the related datastore
            $this->_update_original($this->_data);

            // update the object cache if needed
            $cache and static::$_cached_objects[get_class($this)][static::implode_pk($this->_data)] = $this;

            // mark the object as existing
            $this->_is_new = false;

            // and fire the after-load observers
            $this->observe('after_load');
        }
        else
        {
            // new object, fire the after-create observers
            $this->observe('after_create');
        }
    }
```

This also fixes a cache bug: when using as_object(), $data will be empty. Which means `static::implode_pk($data)` will return null, so the object will not be cached.

keigoi:
Thanks for your quick and great fix:)

I don't intend to bother you :p but `as_object()` costs high if there's vast amount of rows, so you must count rows before querying actual data.

The code would be modified as follows:

``` php
$expr = DB::expr('MATCH (title,body) AGAINST ('.DB::quote($words).' IN BOOLEAN MODE)';

$count = DB::query('SELECT COUNT(*) FROM articles')->where($expr)->execute();

$pagination = Pagination::forge('mydefault', array('total_items' => $count,'name'=>'mydefault'));
$this->template->set_global('pagination', $pagination->render(), false);

$result = DB::select()
    ->from('articles')
    ->where($expr)
    ->limit($pagination->per_page)
    ->offset($pagination->offset)
    ->as_object('Model_Article')
    ->execute();
```

I think this is rather COMMON pattern in FuelPHP. Nevertheless, it became non-modular and hard-to-reuse for other tables. 

See that there's three string constants 'SELECT COUNT(*) FROM articles', 'articles' and 'Model_Article'. If we want to generalize it, you must parameterize on three places.
On the contrary, the version using `custom()` above can be reused for other tables by just replacing Model_Article::query() to a query object of other class.

The points here are that (1) DB::select() is too concrete to handle, and (2) sometimes we must make SQL having additional columns on top of our model columns.
Since we finally get Model_Article object, why don't we go straight to add `custom` data feature into standard \ORM\Query?

WanWizard:
It was an example, obviously not complete code. 

In general, ORM queries are always more expensive than the exact same question using DB calls. So if cost is a concern, you really shouldn't use an ORM.

And have you tested the new constructor? I haven't committed it yet, no time to test it atm.

As to your other remark: if you have the model name, then calling `$model::table()` will get you the table name. 

If you then add a generic static method to your base model class with this code, you can just do something like `Model_Article::paginate(2,25)` which can return `array('pagination' => $paginationobject, 'data' => $queryresult)`.

You can do all that without adding this to the code code. And I prefer it this way, this is quite application specific.

WanWizard:
This can't be merged anymore. Since Fuel v1 is approaching EoL, this should be addressed in the new v2 ORM. If this is still an issue, please open an issue at https://github.com/fuelphp/orm/issues.

sagikazarmark:
IMO this has been discussed several times: ORM is not for things like this. ORM is a tool for fixed data structures. It only adds some extra complexity, also makes things harder to test, etc. My two cents.

WanWizard:
I don't have a problem with this perse, but with the current codebase it isn't possible without hitting this complexity.

PR #337, opened 2014-02-13
Detailing available options in model and query doc blocks.

apdunston:
Added options keys to documentation for model and query to make them more available in documentation. This came from my hunt to find out there was a from_cache option.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -524,7 +524,7 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
      * Find one or more entries
      *
      * @param int|null $id
-     * @param array $options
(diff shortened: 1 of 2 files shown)


emlynwest:
Looks a little messy. Personally this is something I think should be in http://github.com/fuel/docs

apdunston:
I agree that it is messy. I agree that it should also go there. But just a little clue here could save an hour of work for each adopter who stumbles on caching issues.

emlynwest:
This is true but then perhaps just a note in the class' phpdoc would do.

WanWizard:
If it's in the docs, everyone can find it. Nobody should have to dive into the code (ideally ;-) to find out how it works. So I agree it should be in the docs.

apdunston:
I'll close this out for now. If I get time later, I'll write this up in the docs and do a pull request there. Thanks for the feedback, y'all.

PR #339, opened 2014-02-18
Loose check for all int types in Model::is_changed()

Savageman:
Actual code works well when `data_type` is `int`, but does not take into account `tinyint` or `mediumint` or `int unsigned`.

The `type` key from the `$properties` always contain `int` in all this situations.

If this is the correct change, we should also change the doc, which doesn't include the `type` key in the examples.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1672,7 +1672,7 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 			{
 				if (array_key_exists($p, $this->_original))
 				{
-					if (array_key_exists('data_type', $properties[$p]) and $properties[$p]['data_type'] == 'int')
+					if (array_key_exists('type', $properties[$p]) and $properties[$p]['type'] == 'int')
 					{
 						if ($this->{$p} != $this->_original[$p])
 						{


WanWizard:
Not to happy with that, there is a difference between 'type' and 'datatype'.

It would be better to retrieve Observer\Typing::$type_mappings, and do a lookup of 'data_type', they you will catch them all, and also any future data types that may be introduced. You'll probably want loose typing for float's as well...

Savageman:
I can update my pull request to do that.

The only remaining problem is `<*>int unsigned` which won't be found in the `$type_mapping`. How should I solve this one?

WanWizard:
The 'data_type' property of the model column definition is only there for Typing purposes. `int unsigned` is not a valid value, as it is not defined in the Typing observer. 

In Typing terms, it should be defined as `int`, with a "min" value of zero.

Savageman:
It's what we get when using auto-generated properties (we don't write them by hands, we copy/paste auto-generated properties into our models).

Then it's a bug in the script which generates the properties?

WanWizard:
What exactly do you use, the fromdb task? I expect it just pulls in and uses whatever the DB reports, and that doesn't seem to be the correct way of handling it.

Savageman:
1. We write our models without the properties ;
2. We call Model::properties() to retrieve them ;
3. We copy-paste the result into our models.

WanWizard:
And using PDO I assume?

Problem is that it is using DB::list_columns(), which in turn tries to unify the datatype and type settings, which fails for PDO since it has to be RDBMS agnostic, so it doesn't do a particularly good job.

Use the MySQLi driver to generate your properties, if you want correct results.

Savageman:
We don't use PDO. Now we use mysqli, but we previously used mysql that and I believe all columns were generated with the latter one.

But I just tested with MySQLi and I get the same result `data_type = unsigned int` so it doesn't make a difference which driver I'm using.

WanWizard:
Found the issue. the data_type() method in the connection driver lists "int unsigned", not "unsigned int". So this looks like a bug.

I wonder if "unsigned" is always listed first, because currently this is defined:

```
'bigint unsigned'           => array('type' => 'int', 'min' => '0', 'max' => '18446744073709551615'),
'decimal unsigned'          => array('type' => 'float', 'exact' => true, 'min' => '0'),
'double precision unsigned' => array('type' => 'float', 'min' => '0'),
'double unsigned'           => array('type' => 'float', 'min' => '0'),
'fixed unsigned'            => array('type' => 'float', 'exact' => true, 'min' => '0'),
'float unsigned'            => array('type' => 'float', 'min' => '0'),
'int unsigned'              => array('type' => 'int', 'min' => '0', 'max' => '4294967295'),
'integer unsigned'          => array('type' => 'int', 'min' => '0', 'max' => '4294967295'),
'mediumint unsigned'        => array('type' => 'int', 'min' => '0', 'max' => '16777215'),
'numeric unsigned'          => array('type' => 'float', 'exact' => true, 'min' => '0'),
'real unsigned'             => array('type' => 'float', 'min' => '0'),
'smallint unsigned'         => array('type' => 'int', 'min' => '0', 'max' => '65535'),
'tinyint unsigned'          => array('type' => 'int', 'min' => '0', 'max' => '255'),
```

WanWizard:
Just checked, the driver does a "SHOW FULL COLUMNS", and that returns a string like "int(10) unsigned". 

The driver converts that to `array('int unsigned', 10)` which is the correct syntax to be unified using the table above. So the question is why in your case the result is something else...

Savageman:
Sorry, I got `int unsigned` not `unsigned int`.

The original problem still persists since it doesn't do loose typing for such values.

WanWizard:
You shouldn't see that value at all, it should be converted to `array('type' => 'int', 'min' => '0', 'max' => '4294967295')` as per the table shown above. So the question is, why doesn't that happen?

WanWizard:
Think I found it. 'data_type' contains the raw type, 'type' contains the unified type. Which leaves us with an interesting challenge, because both the base model and the typing observer check the 'data_type' and not the 'type'.

Savageman:
Yup, that's why I wanted to use `type`instead of `data_type` in my PR :)

WanWizard:
But that will break BC (all my models that have hand-crafted properties, don't have a type, only a data_type). So both will have to be checked. Commit is already on the way...

Savageman:
I'm happy with the fix you made, will work fine for us. Thanks!

PR #342, opened 2014-03-22
Update slug.php Observer. Uniqueness as option

mikepmtl:
I use slugs very often simply to give the end use a better description of the URL and for SEO.

Ex.  "http://server.com/doc/281/communications/about/about-the-communications-department/"

In a case like this I would prefer not to have uniqueness in the slug as I only require the ID (281) and I dislike slugs with numbers in them (about, about-1, about-2) since many "departments" in this case will have an "About" page.

### DIFF
--- a/classes/observer/slug.php
+++ b/classes/observer/slug.php
@@ -34,10 +34,10 @@ class Observer_Slug extends Observer
 	 */
 	public static $separator = '-';
 
-    /**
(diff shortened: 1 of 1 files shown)


WanWizard:
Can you update the docs too?

mikepmtl:
Sure let me go find the docs.  :)

PR #344, opened 2014-05-09
Update slug.php

sagikazarmark:
Bad logical operator

### DIFF
--- a/classes/observer/slug.php
+++ b/classes/observer/slug.php
@@ -93,7 +93,7 @@ class Observer_Slug extends Observer
 	public function before_insert(Model $obj)
 	{
 		// slug should be overwritten if it is enabled to be or there is no manually assigned value
-		$overwrite = $this->_overwrite === true or empty($obj->{$this->_property});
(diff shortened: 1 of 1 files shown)


emlynwest:
This seems pretty pointless?

sagikazarmark:
No it is not.

``` php
// Evaluates false
$overwrite = false or true;

// Evaluates true
$overwrite = false || true;
```

emlynwest:
I see now. My mistake!

okhayat:
Indeed. Is this a bug in PHP? Shouldn't OR act just like ||?!

sagikazarmark:
@okhayat No, it isn't.

okhayat:
But OR and || should be the same:
http://php.net/manual/en/language.operators.logical.php
$a or $b    Or  TRUE if either $a or $b is TRUE.
$a || $b    Or  TRUE if either $a or $b is TRUE.

sagikazarmark:
@okhayat You are wrong. Read the page you linked more carefully. (Hint: Operator precedence)

okhayat:
Yes. Operator precedence is confusing sometimes.
Good catch.

PR #346, opened 2014-05-11
Update query.php

codeChap:
Fix for "Undefined offset error" from the orm auth package.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -290,7 +290,7 @@ class Query
 		}
 		foreach ($fields as $val)
 		{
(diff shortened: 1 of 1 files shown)


WanWizard:
I wonder what caused [0] not to exist? What case caused it to be an empty array (or an assoc array that does not have a [0] key)?

sagikazarmark:
It is, but I am not sure it would be the best solution. The empty array case should be handled globally, I am sure there are some other cases where it exists.

codeChap:
\Auth::check() was possibly causing it, screen grab here: http://cl.ly/image/0K1r3O132z0v

PR #352, opened 2014-06-10
In "count" queries, do not add all the columns from relations.

sjamaan:
This can make the difference between using no indexes and writing to disk for a
temp table and bringing down the system versus using indexes and completing in
a few milliseconds.

The comment "// Remove the current select and" seems to imply that this was the
original idea, but maybe the API somehow got changed which caused a performance
regression without anyone noticing it.

PS: I am unsure what to do with the subselect situation, so I decided to keep it the same
for fear of breaking things. If you have some code that uses this extensively I can
recommend trying to fix that too, it may make a huge difference in performance.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -715,7 +715,7 @@ class Query
 	 *
 	 * @param   \Fuel\Core\Database_Query_Builder_Where  DB where() query object
 	 * @param   array $columns  Optionally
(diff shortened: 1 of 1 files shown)


WanWizard:
Can you close this, and send the PR on 1.8/develop?

PR #353, opened 2014-06-11
In "count" queries, do not add all the columns from relations.

sjamaan:
Same as https://github.com/fuel/orm/pull/352, but cherry-picked into 1.8:

This can make the difference between using no indexes and writing to disk for a
temp table and bringing down the system versus using indexes and completing in
a few milliseconds.

The comment "// Remove the current select and" seems to imply that this was the
original idea, but maybe the API somehow got changed which caused a performance
regression without anyone noticing it.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -728,7 +728,7 @@ class Query
 	 *
 	 * @param   \Fuel\Core\Database_Query_Builder_Where  DB where() query object
 	 * @param   array $columns  Optionally
-	 * @param   string $type    Type of query to build (select/update/delete/insert)
+	 * @param   string $type    Type of query to build (count/select/update/delete/insert)
 	 *
 	 * @throws \FuelException            Models cannot be related between different database connections
 	 * @throws \UnexpectedValueException Trying to get the relation of an unloaded relation
@@ -737,6 +737,8 @@ class Query
 	 */
 	public function build_query(\Fuel\Core\Database_Query_Builder_Where $query, $columns = array(), $type = 'select')
 	{
+		$read_query = in_array($type, array('select', 'count'));
+		
 		// Get the limit
 		if ( ! is_null($this->limit))
 		{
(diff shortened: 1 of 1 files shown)


WanWizard:
@stevewest any comments on this?

emlynwest:
As always I am a little unsure when looking at making changes to v1 orm. The code is so messy I can hardly understand what the internals are doing. Given that v2 orm is underway I'd say leave this be for now and if you are worried about performance then use the DB class and get the table name from your model.

WanWizard:
My problem exactly, I think this is change is ok, but I'm worried about the side effects.

emlynwest:
I would say wait for the next 1.7 release then merge it after that.

WanWizard:
Good plan.

emlynwest:
@WanWizard I've had reports that this merge has caused where statements to not be applied to count queries, so seems there where some unintended side effects cause by this.

WanWizard:
As far as I can see this is only about selecting columns for the result, it is not a problem to have `SELECT a FROM table WHERE b = something`, WHERE clauses do not require the column to be present in the result.

Can you ask for examples, like a profiler dump of the count query, so we can see what and where it goes wrong? Reverting this one might be complex.

andrewwheal:
We are patching a couple of our sites. After that I will look into simple examples and have a look into what might be happening.

andrewwheal:
I have made a simplified example controller which can be seen here - https://gist.github.com/SicoAnimal/bfbe5fc78e52bdb8a5e9

Here is the output of the above from before and after this PR was merged - https://gist.github.com/SicoAnimal/38a515ec9af9011ab49a

andrewwheal:
The issue was found with a local override of the Orm\Query class which had a copy+paste of the build_query() method with a small custom addition.

Testing without this override shows this PR to be fine.

WanWizard:
Happy to hear that, thanks for the feedback.

PR #354, opened 2014-06-12
Fixed bug in is_changed method to unset correct pk.

MacTwister:
Also if model or any relation is_new, then can return true.

I hope this is correct. This worked for me, as before is_changed was always returning TRUE even if nothing changed.


WanWizard:
Any feedback on this issue?

PR #364, opened 2014-12-19
Workaround for limitation of PHP 5.x COW design

pmccarren:
When pulling a large number of records the ORM would get
exponentially slower. The issue was tracked down to a PHP
limitation. We filed https://bugs.php.net/bug.php?id=68624

PHP developer suggests issue will be resolved in PHP 7

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -124,6 +124,11 @@ class Query
 	protected $from_cache = true;
 
 	/**
+	 * @var  mixed  hydration result implemented to prevent https://bugs.php.net/bug.php?id=68624
+	 */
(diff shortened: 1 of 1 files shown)


WanWizard:
@stevewest any objections?

emlynwest:
I don't have an issue with this, it's certainly not the worst workaround that I have seen and if it can make the ORM perform better then I'm happy for it to be merged.

WanWizard:
Had to revert it, this change made results global, which completely f'd up the results if hydrate() had to iterate (which is does when eager loading relations).

PR #371, opened 2015-03-09
Support relations select condition.

n0ts:
Support like this.

$post = Model_Post::query()
    ->related('articles', array('select' => array('article_id', 'active', 'published')))
    ->where('articles.user.active', '=', 1)
    ->order_by('articles.published', 'desc')
    ->get_one();


emlynwest:
I don't think this should be merged, it encourages partial selects that are known to have confusing side effects with the current orm's cache.

n0ts:
@stevewest Thanks for your comment. Sorry I don't think about orm's model cache. Thanks.

PR #373, opened 2015-04-05
Avoid to destroy the overridden settings

takayukioda:
Current orm will override settings by empty array, when $_properties were something like this.

``` php
$_properties = [
    0 => 'id',
    'id' => ['data_type', 'int'],
];
```

will became

``` php
$_properties = [
    'id' => [],
];
```

Order of the definition won't help to avoid this problem.

By this PR, orm will no longer destroy any overridden settings.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -316,6 +316,10 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 				if (is_string($p))
 				{
 					unset($properties[$key]);
+					if (isset($properties[$p]) and ! empty($properties[$p]))
+					{
+						continue;
+					}
 					$properties[$p] = array();
 				}
 			}


emlynwest:
The definition of the properties you show there is incorrect. What you are describing is expected behaviour. If you want to avoid it then you can define the empty properties as an empty array:

``` php
$_properties = [
    'id' => [],
]
```

takayukioda:
@stevewest Thank you for the reply and I understand your opinion.
Though, some packages such as [\Auth\Model\Auth_User](https://github.com/fuel/auth/blob/1.8/develop/classes/model/auth/user.php#L38) have defined their properties not as an empty array but as a string.
For those patterns, I think there is no way to override settings. Or is there any way to do so?

PR #377, opened 2015-05-16
Extended hasmany and manymany "get" method by handling "limit" condition

feldsam:

### DIFF
--- a/classes/hasmany.php
+++ b/classes/hasmany.php
@@ -75,6 +75,11 @@ class HasMany extends Relation
 				$query->order_by($field, $direction);
 			}
 		}
+		
+		if($limit = \Arr::get($conditions, 'limit'))
+		{
+			$query->limit($limit);
+		}
 
 		return $query->get();
 	}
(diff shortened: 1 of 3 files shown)


emlynwest:
Looks ok but could you also add documentation for this over at https://github.com/fuel/docs please?

WanWizard:
This the docs also need a big warning: this means results from this relation might be incomplete due to the limit, and this can have implications because the parent object is cached.

Before accepting this PR, we have to make sure that if you define two indentical relations, one with limit, one without, and you fetch the one with the limit first, the other one still produces the correct results (and I fear not).

emlynwest:
Ah right, I had somehow forgotten about the caching.

feldsam:
Hi, when you use already implemented where condition them you also get different result set. So?

WanWizard:
As said, if it doesn't interfere, it's not a problem. Just needs a check, to make sure there are no compatibility issues created.

feldsam:
I think that, if where condition doesn't interfere then limit condition doesn't interfere too.

WanWizard:
As Morpheus said to Neo: "Don't think you are, know you are". ;-)

PR #382, opened 2015-07-14
Support for custom names of PK column

ppatrik:
tables should have other PK names than 'id', eg. 'table_id'

### DIFF
--- a/classes/model/temporal.php
+++ b/classes/model/temporal.php
@@ -151,7 +151,7 @@ class Model_Temporal extends Model
 		self::disable_primary_key_check();
 
 		$query = static::query()
-			->where('id', $id)
+			->where(static::getIdColumn(), $id)
 			->where($timestamp_start_name, '<=', $timestamp)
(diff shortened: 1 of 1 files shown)


mikepmtl:
It already is supported.

protected static $_primary_key = array('aid');

http://fuelphp.com/docs/packages/orm/creating_models.html

ppatrik:
yes but here in temporal model it was written to only accept `id`
see

classes/model/temporal.php line 153

```
        $query = static::query()
            ->where('id', $id)    <------- this is not correct
            ->where($timestamp_start_name, '<=', $timestamp)
            ->where($timestamp_end_name, '>', $timestamp);
```

WanWizard:
You'll have to wait until @stevewest gets back, to asses it.

I'm pretty sure this has something to do with the complexities of compound PK's, something your PR doesn't address either.

emlynwest:
Temporal is limited to "id" by design. It's a current limitation of the v1 temporal model and will be addressed in v2.

PR #383, opened 2015-08-25
Add a method to get the primary key's as an array

xploSEoF:
This creates an array containing the relevant information for identification of the object. This could be used in many locations, specifically in my case, logging.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1271,13 +1271,14 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 	 */
 	public function get_pk_array()
 	{
-	    $array = array_flip(static::primary_key());
+		$array = array_flip(static::primary_key());
 
-		foreach($array as $key => &$value) {
+		foreach($array as $key => &$value)
+		{
 			$value = $this->get($key);
 		}
 
-        return $array;
+		return $array;
 	}
 
 	/**
--- a/classes/model.php
+++ b/classes/model.php
@@ -1263,6 +1263,22 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 
 		return $this;
 	}
+	
+	/**
+	 * Provide the identifying details in the form of an array
+	 * 
+	 * @return array
+	 */
+	public function get_pk_array()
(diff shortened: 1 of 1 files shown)


emlynwest:
Functionality looks fine but please fix the coding style. Indent with tabs and `{}` on new lines please. Also possibly the function would be better named `get_pk_values()` or something akin to that to better describe what it does.

WanWizard:
I'd say `get_pk()` would be enough, no need to add "array" or "values" to it? I'd only do that if you have or intend to have more methods starting with `get_pk`? 

Also, please update the documenation with the new method.

emlynwest:
Only issue is that there's already a "get pk" method that returns the property names.

WanWizard:
Then "values" seems better than "array", agreed.

xploSEoF:
Or assoc, since values implies it is literally just the values...

emlynwest:
I'd be happy with either `assoc` or `values`, both work for me.

xploSEoF:
Where would you like documentation to be added? I can't find any of the functions on \Orm\Model in any of the documentation, with the exception of the very basic ones on the CRUD page.

WanWizard:
You got a point, it does miss a method reference, so nowhere to add it...

xploSEoF:
Was this not merged then, or has GitHub not identified the merge?

WanWizard:
No, it wasn't. And it was a PR against a now outdated branch, so it couldn't be merged anymore.

WanWizard:
Also, I wonder if that code (in particular the value reference) still works in PHP 7, as it always loops over a copy now.

@stevewest can you comment?

xploSEoF:
Given my understanding over the PHP7 changes, this should continue to work.

Foreach loops in PHP < 7 loops over a copy, but with references when necessary. PHP >= 7 loops through the array itself.

WanWizard:
Ok, then creating a new PR for 1.9/develop should not be a problem.

Regarding the method name, @stevewest stated `get_pk()` already existed, but I can''t find it?

PR #385, opened 2015-09-23
Workaround for limitation of PHP 5.x COW design (Revive)

sj-i:
Reviving #364
This problem can be workarounded also by wrapping the result array in an object instance.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -1116,16 +1116,16 @@ class Query
 	/**
 	 * Hydrate model instances with retrieved data
 	 *
-	 * @param   array   &$row   Row from the database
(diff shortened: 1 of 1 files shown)


WanWizard:
Manually merged it into one of our apps. After an hour of testing I can't find a fault, so I pushed it to the production server for this app. I'd like to make sure it has no side effects before we merge this.

@stevewest ok with you?

emlynwest:
Nothing jumps out at me that looks like it is going to cause issues. I'd say merge it in, we can always roll it back if issues start to occur.

WanWizard:
Fine by me.

PR #387, opened 2015-12-16
SQlServer, Oracle

atabak:
More Databse Support Like SQLServer, Oracle


WanWizard:
Why do you want to push old 1.7 code to 1.8 what already had that code?

If you have an issue or a feature request, you open an issue, not a pull request. And in this case, open it on fuel/core as that provides the DB layer, not the ORM package.

PR #393, opened 2016-07-23
Added purge_before and purge_after events to soft model

codeChap:
Needed to make it easier to permanently remove certain files after or before a purge.


classes/model/soft.php
```diff
@@ -180,11 +180,14 @@ protected function delete_self()
 	 */
 	public function purge($cascade = null, $use_transaction = false)
 	{
+    $this->observe('before_purge');
```
emlynwest: Looks like indentation is incorrect here.

PR #395, opened 2016-10-13
Fixed WHERE condition doesn’t work when use getone in the soft_filter

linzhengen:
# doesn’t work by these case

PHP logic

``` php
 // case 1
 Model_Project::find('first',
  array(
      'related' => array('project_users'),
      'where' => array(
          array('project_users.id', '=', 2),
          array('code', '=', 'code1'),
      ),
  ));

 // case 2
 Model_Project::query(
  array(
      'related' => array('project_users'),
      'where' => array(
          array('project_users.id', '=', 2),
          array('code', '=', 'code1'),
      ),
  ))->get_one();

 // case 3
 Model_Project::query()
    ->related('project_users')
    ->where('project_users.id', '=', 2),
    ->where('code', '=', 'code1'),
    ->get_one();
```

**generated the same SQL**
The `t0` subquery doesn’t work, because it don't have any where condition for `limit 1`

``` sql
SELECT 
  `t0`.`id` AS `t0_c0`, 
  `t0`.`code` AS `t0_c1`, 
  `t0`.`deleted_at` AS `t0_c2`, 
  `t0`.`created_at` AS `t0_c3`, 
  `t0`.`updated_at` AS `t0_c4`, 
  `t1`.`id` AS `t1_c0`, 
  `t1`.`project_id` AS `t1_c1`, 
  `t1`.`user_id` AS `t1_c2`, 
  `t1`.`deleted_at` AS `t1_c3`, 
  `t1`.`created_at` AS `t1_c4`, 
  `t1`.`updated_at` AS `t1_c5` 
FROM 
  (
    SELECT 
      `t0`.`id`, 
      `t0`.`code`, 
      `t0`.`deleted_at`, 
      `t0`.`created_at`, 
      `t0`.`updated_at` 
    FROM 
      `projects` AS `t0` 
    WHERE 
      `t0`.`deleted_at` IS null 
    ORDER BY 
      `t0`.`id` ASC 
    LIMIT 
      1
  ) AS `t0` 
  LEFT JOIN `project_users` AS `t1` ON (
    `t0`.`id` = `t1`.`project_id` 
    AND `t1`.`deleted_at` IS NULL
  ) 
WHERE 
  (
    `t1`.`id` = 2 
    AND `t0`.`code` = 'code1' 
  ) 
ORDER BY 
  `t0`.`id` ASC
```
# But it works by these case

```
// case 1
 Model_Project::find('first',
     array(
         'related' => array('project_users' => array(
             'where' => array(array('id', '=', 2))
         )),
         'where' => array(
             array('code', 'code1'),
         ),
     )
 );
// case 2
 Model_Project::query(
     array(
         'related' => array('project_users' => array(
             'where' => array(array('id', '=', 2))
         )),
         'where' => array(
             array('code', 'code1'),
         ),
     )
 ))->get_one();
```

### DIFF
--- a/classes/query/soft.php
+++ b/classes/query/soft.php
@@ -94,21 +94,7 @@ class Query_Soft extends Query
 	{
 		if ($this->_col_name !== null)
 		{
-			// Capture any filtering that has already been added
-			$current_where = $this->where;
-
-			// If there is no filtering then we don't need to add any special organization
-			if ( ! empty($current_where))
-			{
-				$this->where = array();
-
-				// Make sure the existing filtering is wrapped safely
-				$this->and_where_open();
-				$this->where = array_merge($this->where, $current_where);
-				$this->and_where_close();
-			}
-
-			// Finally add the soft delete filtering
+			// Add the soft delete filtering
 			$this->where($this->_col_name, null);
 		}
 	}


WanWizard:
We don't accept PR's on release branches, only on the latest develop branch.

And the ORM is not a query builder. This behaviour is by design, if you select objects with relation, it will apply the limit on the parent, but it always fetch all children, otherwise the resultset is not consistent. The ORM will never select partial relation data.

If you want your behaviour, you need to use 'rows_limit' instead of 'limit'. This behaviour is documented.

linzhengen:
Thank you for your comment.

PR #401, opened 2016-12-08
Bugfix diff

xploSEoF:
Reopen #396 

Fixes a couple of issues with the difference engine.

Current issues:
- showing "null" for both values when showing the difference for a singular relation - specifically when adding a relation where there wasn't one before
- strict comparison of values for boolean fields (matching type)

These now:
- show the difference exactly for a singular relation when adding a relation where there wasn't one before
- loose comparison of values for boolean fields (not matching type)

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -1816,12 +1816,13 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 			$rel = static::relations($key);
 			if ($rel->singular)
 			{
-				$new_pk = null;
+				$new_pk = empty($val) ? null : $val->implode_pk($val);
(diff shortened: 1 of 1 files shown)


classes/model.php
```diff
@@ -1718,15 +1718,16 @@ public function is_changed($property = null)
 		$properties = static::properties();
 		$relations = static::relations();
 		$property = (array) $property ?: array_merge(array_keys($properties), array_keys($relations));
+		$simple_data_types = ['int','bool'];
```
WanWizard: This is not PHP 5.3.3 compatible, please use `array()` instead.

PR #402, opened 2016-12-08
Add a method to get the primary key's as an associative array

xploSEoF:
Re-open #399

Provides the primary key in a specifically defined array that should be all that is necessary to reverse-lookup the item

This is the effective rebase of the work done for #383

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -2170,6 +2170,22 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 	}
 
 	/**
+	 * Provide the identifying details in the form of an array
+	 *
+	 * @return array
+	 */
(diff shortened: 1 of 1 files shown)


classes/model.php
```diff
@@ -2170,6 +2170,22 @@ public function to_array($custom = false, $recurse = false, $eav = false)
...
 	/**
+	 * Provide the identifying details in the form of an array
+	 *
+	 * @return array
+	 */
+	public function get_pk_assoc()
+	{
+		$array = array_flip(static::primary_key());
+
+		foreach($array as $key => &$value) {
```
WanWizard, line 2181: Coding standards...

(space after foreach, { on a new line)

PR #409, opened 2017-06-21
Fix substr on unicode string

Seb-C:
This replaces `substr` by `mb_substr`.

My specific case happened with a `varchar(250)` field, and a string that was longer. The 250th and 251th bytes of the string were an UTF-8 non-breakable space (0xC2A0).

Since C2 is a control character that can't exist alone, this made everything using this string crash thereafter (including debugging tools).

### DIFF
--- a/classes/observer/typing.php
+++ b/classes/observer/typing.php
@@ -261,9 +261,9 @@ class Observer_Typing
 		if (array_key_exists('character_maximum_length', $settings))
 		{
 			$length  = intval($settings['character_maximum_length']);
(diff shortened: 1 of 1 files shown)


WanWizard:
That code will fail if mbstring isn't loaded. Ideally, this should use \Str::truncate(), which is multibyte aware.

I think this needs to be looked at in more broader terms, quite a bit of the Fuel was written when overloading string functions was the norm, which isn't the case anymore, and is even deprecated as of PHP 7.2.

Issue #415, opened 2017-11-27
Where statement omitted operator problem


yama:
```
->related('author')
->where('author', '=', 1);
```
Work. It's no problem.

```
->related('author')
->where('author', 1);
```
PHP notice display an error. It does not work in the my dev-environment.

https://github.com/fuel/orm/blob/1.8.0.1/classes/hasone.php#L97-L107
This problem can be fixed within this block.

https://github.com/fuel/orm/blob/1.8.0.1/classes/query.php#L526-L529
Add this process.

thanks for fuelphp.

PR #416, opened 2017-11-27
#415 Fix - Where statement omitted operator problem

yama:
https://github.com/fuel/orm/issues/415
PHP notice error occurs if operator is omitted in where statement
meaning equal.


WanWizard:
Thanks!

PR #421, opened 2018-04-28
1.9.1/master

joaopschiazza:
Fix


WanWizard:
You haven't followed the coding standards.

And what is this supposed to fix? Please use an informational commit message, so it is clear in the git log what this commit is about.

PR #431, opened 2020-08-11
fix getting relation data with multiple pks

ysshir:


WanWizard:
That looks like a very dumb error, thanks for fixing it !

PR #433, opened 2020-09-06
Issue with relation without records

AdamSGit:
This PR fix the issue talked about in [this forum thread](https://fuelphp.com/forums/discussion/15290/bug-reportfueldev-1-9i-could-not-get-related-data-using-orm#Item_6) about relations with no records not being initialized, triggering relation lazy-loading.

### DIFF
--- a/classes/model.php
+++ b/classes/model.php
@@ -2083,7 +2083,7 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 				{
 					$this->_data_relations[$property] = $rel->singular ? null : array();
 
-					// If value is null or [], this mean it was a relation without record. Continue, then.
+					// If value is an empty array, this mean it was a relation without record. Continue, then.
 					if ( ! $value)
 					{
 						continue;
--- a/classes/model.php
+++ b/classes/model.php
@@ -2077,9 +2077,17 @@ class Model implements \ArrayAccess, \Iterator, \Sanitization
 			elseif (array_key_exists($property, static::relations()) and is_array($value))
 			{
 				$rel = static::relations($property);
+
+				// Make sure the property is set in the relation array
 				if ( ! isset($this->_data_relations[$property]))
 				{
 					$this->_data_relations[$property] = $rel->singular ? null : array();
+
+					// If value is null or [], this mean it was a relation without record. Continue, then.
+					if ( ! $value)
+					{
+						continue;
+					}
 				}
 
 				if ($rel->singular)
--- a/classes/query.php
+++ b/classes/query.php
@@ -1524,6 +1524,9 @@ class Query
 				continue;
 			}
 
+			// If the current model is an empty relation, still need to initialize it empty later
+			$is_relation_without_records = false;
+
 			// construct the PK string representation for this record
 			if (count($model['pk']) == 1)
 			{
@@ -1531,7 +1534,12 @@ class Query
 				// skip empty results
 				if (is_null($pk))
 				{
-					continue;
+					if( ! $model['relation']) 
+					{
+						continue;
+					}
+
+					$is_relation_without_records = true;
 				}
 			}
 			else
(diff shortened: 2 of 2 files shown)


classes/model.php
```diff
@@ -2077,9 +2077,17 @@ public function from_array(array $values)
 			elseif (array_key_exists($property, static::relations()) and is_array($value))
 			{
 				$rel = static::relations($property);
+
+				// Make sure the property is set in the relation array
 				if ( ! isset($this->_data_relations[$property]))
 				{
 					$this->_data_relations[$property] = $rel->singular ? null : array();
+
+					// If value is null or [], this mean it was a relation without record. Continue, then.
```
WanWizard: value can't be null, because of the `is_array()` check in L#2077

classes/query.php
```diff
@@ -1608,6 +1628,15 @@ public function process_row($row, $models, &$result)
 					}
 				}
 
+				// If this is a relation without records, set it to an empty array and continue
+				// Also check for [] because a previous loop iteration may have already set it up
+				if ($is_relation_without_records and (! isset($target[$current]) or $target[$current] === []) )
+				{
+					$target[$current] = [];
```
WanWizard: Please use `array()` instead of `[]`, for compatbility reasons.

classes/query.php
```diff
@@ -1524,14 +1524,22 @@ public function process_row($row, $models, &$result)
 				continue;
 			}
 
+			// If the current model is an empty relation, still need to initialize it empty later
+			$is_relation_without_records = false;
```
WanWizard: I have a bit of an issue with this double negative. Technically ok, but imho more complex to read than `$has_related_records = true`.

classes/query.php
```diff
@@ -1546,37 +1554,49 @@ public function process_row($row, $models, &$result)
...
+				{
+					$pk = '['.implode('][', $pks).']';
 				}
-				$pk = '['.implode('][', $pks).']';
 			}
 
-			// do we need to check for cached objects
-			if ($this->from_cache)
+			// This part of code shouldn't run if the current model is a relation with no record(s)
+			if( ! $is_relation_without_records )
```
WanWizard: Same here,
````
if ( $has_related_records)
````
is a lot easier to read.

File: classes/query.php [hunk 5]
```diff
@@ -1546,37 +1554,49 @@ public function process_row($row, $models, &$result)
...
+					if ($obj)
 					{
-						if (empty($obj->{$column}))
+						// update the object with fields missing from it
+						$new = array();
+						foreach ($record as $column => $value)
 						{
-							$obj->{$column} = $value;
-							$new[$column] = $value;
+							if (empty($obj->{$column}))
```
WanWizard, line 1585: I know that this is existing code, but reading it again, I wonder if this goes ok. 

The idea here is that if the object in cache was created by a partial select, missing colums to the cached object will be added here. But what if `$obj->{$column}` is `int(0)` or `string('')`? I can't test it now, but I wonder if this shouldn't be an `isset()` instead of an `empty()` check...

WanWizard: I know that this is existing code, but reading it again, I wonder if this goes ok. 

The idea here is that if the object in cache was created by a partial select, missing colums to the cached object will be added here. But what if `$obj->{$column}` is `int(0)` or `string('')`? I can't test it now, but I wonder if this shouldn't be an `isset()` instead of an `empty()` check...

classes/query.php
WanWizard: Also, the missing relation could be a "one" or a "many", so I wonder if that should be checked here, because singular relations shouldn't be initialized with an empty array, but with null?

AdamSGit: I through the same, but since in the model the check for relation is only done for array(), I wouldn't modify too much this part of the code to avoid causing unexpected regressions.

AdamSGit:
I applied your reviews, except for the one I answered.

classes/query.php
WanWizard: Not sure that is true.

In https://github.com/fuel/orm/blob/1.9/develop/classes/model.php#L1220 it will check if related data exists by means of an `array_key_exists` (which will be true after this code change),  and it then procedes to return the value.

Now, application code will for a HasOne relation expect the value to be either `null` (no related object) or an instance of `\Orm\Model`. But not an array. So although the ORM itself won't be too bothered with it, it is possible that user code will fail.

So I think it's probably better to replace it by
````
$target[$current] = $model['singular'] ? null : array();
````

Comment (AdamSGit) [reply classes/query.php]: I wrote it that way in the first place, but the line 2077 of model.php in `from_array` method used in query.php, which is : 
```
elseif (array_key_exists($property, static::relations()) and is_array($value))
```
Check that the value is an array to enter the code instruction block that handle the relation part when building model. So, your proposal would require to change that line for : 
```
elseif (array_key_exists($property, static::relations()) and (is_array($value) or $value === null))
```
I didn't do it to avoid causing regressions, even if there shouldn't be any issue. Tell me what you prefer.

Comment (WanWizard) [reply classes/query.php]: The `$value` used there comes from the argument `$values` of `from_array()`, so not related to this issue.

For converting array data into model objects, it should always be an array, either an array of values (for singlar objects) or an array of arrays (for non-singular objects).

So I'd say your first hunch was correct ;-).

Comment (WanWizard) [reply classes/query.php]: p.s. where in `query.php` is `from_array()` used? Can't seem to find it.

Comment (AdamSGit) [reply classes/query.php]: Trace : 
**Query.php -> get() -> l.1469** 
```
$objects[$key] = $this->model::forge($record, false, $this->view ? $this->view['_name'] : null, static::$caching);
```
**Model.php -> forge() -> l.145** 
```
return new static($data, $new, $view, $cache); // Call constructor
```
**Model.php -> __construct() -> l.923** 
```
$this->from_array($data);
```
**Model.php -> from_array() -> l.2077** 
```
elseif (array_key_exists($property, static::relations()) and is_array($value))
```
My first hunch was invalidated by tests, forcing me to trace back why singular relations set to null instead of [] weren't defined. The last line should be modified as given in my last comment for it to work.

WanWizard: Hmm... Bloody difficult if all you have is a webpage with code, and no system to actually test on... :disappointed: 

I get it now. The result of the hydration is fed into `Model::forge()` which in turn calls `from_array()` to create the model objects. Which does indeed requires arrays, the correct initialisation is done on L#2082.

Which means no more points of discussion, this can be merged, right?

AdamSGit: Right. :-)

PR #437, opened 2020-09-09
Fix issue with nested relationship

AdamSGit:
I also noticed that, in case of nested relationship, recents changes would create an empty model with an empty key for the parent relation, because of the lack of check if target is not null. This pull request fix it.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -1621,7 +1621,7 @@ class Query
 					{
 						$target =& $target[$parent];
 					}
-					else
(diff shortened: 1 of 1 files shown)


WanWizard:
Are you sure you mean `else if` and not `elseif`, as they will behave differently if more is changed.

I prefer it without the space, it makes it clear what was meant.

AdamSGit:
Sure thing.

PR #438, opened 2020-09-11
fix loading for singular relation's relations

ysshir:
This PR fix is for 1st problem of #435.

Consistency problem is quite big problem.
but I would like to resolve at least this one.

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -1492,6 +1492,9 @@ class Query
 		// storage for this rows pk's
 		$rowpks = array();
 
+		// keep singular or not
+		$is_singular = array();
+
 		// process the models in the row
 		foreach ($models as $alias => $model)
 		{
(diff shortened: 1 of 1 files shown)


WanWizard:
This issue should be fixed in https://github.com/fuel/orm/commit/13aa57b09247cfe41d5d781932ff5afb0a529446

ysshir:
I think it's not fixed yet.

still not fetched in get method.

the problem is Line:1625
```
if ($target[$parent] instanceOf Model)
```

there is no Model instance, except the Model is cached and use from_cache.
The GrandChild data is stored in $_custom_data, not _data_relations.

and GrandChildren data are loaded lazily.

WanWizard:
You are right, the dissussion about this issue goes on here: https://github.com/fuel/orm/pull/435

PR #447, opened 2021-03-11
error fix

hotlabs:
Undefined array key ""
PKGPATH/orm/classes/query.php @ line 1563

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -1549,7 +1549,7 @@ class Query
 				// store it, but do not overwrite an existing result!
 				if ($model['singular'] === true)
 				{
-					if ( ! array_key_exists($current, $pointers[$parent]))
+					if (!isset($pointers[$parent]) || ! array_key_exists($current, $pointers[$parent]))
 					{
 						$pointers[$parent][$current] = $record;
 					}
@@ -1560,7 +1560,7 @@ class Query
 				}
 				else
 				{
-					if ( ! array_key_exists($current, $pointers[$parent]))
+					if (!isset($pointers[$parent]) || !array_key_exists($current, $pointers[$parent]))
 					{
 						$pointers[$parent][$current] = array($pk => $record);
 					}


WanWizard:
Can you give me an example in which a related record doesn't have a parent? 

What was `$pointers`, `$current` and `$parent` at that point?

hotlabs:
error when using auth package, Ormauth,
	function Auth::verified()
			
        $parent = NULL
        $pointers = empty erray()
        $current = (str)'metadata'
php 8.0.3

WanWizard:
and what is `$current` at that point? 

Because that suggests a first level related object, for root objects both `$current` and `$parent` should be `NULL`, and in that case, `$pointers[NULL]` should be created first and should therefore exist.

hotlabs:
> и что `$current`в этот момент?
> 
> Поскольку это предполагает объект, связанный с первым уровнем, для корневых объектов оба `$current`и `$parent`должны быть `NULL`, и в этом случае `$pointers[NULL]`должны быть созданы в первую очередь и поэтому должны существовать.

    $current = (str)'metadata'

WanWizard:
There is only one query in Ormauth that uses metadata:
````
				// find the user
				$this->user = \Model\Auth_User::query()
					->select(\Config::get('ormauth.table_columns', array()))
					->related('metadata')
					->where('username', '=', $username)
					->get_one();
````
which is a very simple and often occuring query. I'll see if I can reproduce it, because this error shouldn't happen.

hotlabs:
Без названия

hotlabs:
this error after install fuel and include auth package - without changing the code

WanWizard:
I can reproduce it, looking at it now.

hotlabs:
Yes, please fix this

WanWizard:
Found the problem, fix is on the way, testing it now.

WanWizard:
Fixed in 18dfbdeef0432c7de0bf2fb60bb44fd366785b90

PR #448, opened 2021-04-12
fix select with alias column

ysshir:
Previous PR https://github.com/fuel/orm/pull/432 doesn't work right now.

sample code.
```
class Model_Temp extends Orm\Model {

    public static function _init() {
        static::$_table_name = $table = 'temp_' . time();

        DB::query(<<<SQL
                CREATE TEMPORARY TABLE {$table} (
                    `id`            int(11) not null auto_increment,
                    `value`         varchar(255) null,
                    primary key (`id`)
                );
                SQL
        )->execute();
    }

    protected static $_table_name  = null;
    protected static $_primary_key = ['id'];
    protected static $_properties  = [
        'id',
        'value',
    ];
}


Model_Temp::forge(['value' => 10])->save();
Model_Temp::forge(['value' => 20])->save();
Model_Temp::forge(['value' => 30])->save();
Model_Temp::flush_cache();

$data = Model_Temp::query(['select' => [[DB::expr('sum(value)'), 'total']]])
                  ->get_one();


echo $data->total;
```

### DIFF
--- a/classes/query.php
+++ b/classes/query.php
@@ -527,7 +527,14 @@ class Query
 			$out = array();
 			foreach($this->select as $k => $v)
 			{
-				$out[] = is_array($v) ? array($v[1], $k) : array($v, $k);
+				if (is_array($v))
+				{
+					$out[$v[1]] = array($v[0], $k);
+				}
+				else
+				{
+					$out[] = array($v, $k);
+				}
 			}
 
 			// set select back to before the PKs were added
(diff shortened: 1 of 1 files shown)


ysshir:
I needed to think about selecting DB::expr() without alias.
so sorry...

```
$data = Model_Temp::query(['select' => [DB::expr('sum(value)'), [DB::expr('sum(value)'), 'total']]])
                  ->get_one();

```

